PR #11, opened 2023-02-21
chore: Fix exit handling test suite

gnarea:
As in, make sure all tests pass. ESLint is reporting errors on these files.

### DIFF
--- a/src/utilities/exitHandling.spec.ts
+++ b/src/utilities/exitHandling.spec.ts
@@ -13,7 +13,9 @@ beforeEach(() => {
 });
 
 const mockProcessOn = mockSpy(jest.spyOn(process, 'on'));
-const mockProcessExit = mockSpy(jest.spyOn(process, 'exit'));
+const mockProcessExit = mockSpy(jest.spyOn(process, 'exit'), () => {
+  // Do nothing
+});
 
 describe('configureExitHandling', () => {
   beforeEach(() => {


src/utilities/exitHandling.spec.ts
```diff
@@ -13,7 +13,9 @@ beforeEach(() => {
 });
 
 const mockProcessOn = mockSpy(jest.spyOn(process, 'on'));
-const mockProcessExit = mockSpy(jest.spyOn(process, 'exit'));
+const mockProcessExit = mockSpy(jest.spyOn(process, 'exit'), () => {
+  // Do nothing
```
gnarea, line 17: Found the fix here: https://stackoverflow.com/a/51448240/129437

annaavanesyan: I would be better if we omit comments, default eslint configuration is forbidding those as well [no-inline-comments]

gnarea: If I omit the comment, that'd violate another rule:

Screenshot_20230221_150928

Note that this isn't an inline comment though.

annaavanesyan: approved :)

gnarea: Thanks! Will merge

PR #17, opened 2023-02-22
chore: Fix various ESLint violations

gnarea:
- Exit handling.
- Logging test utils.
- `env-var` test utils.
- Fastify utils.


src/utilities/exitHandling.ts
```diff
@@ -4,6 +4,6 @@ export function configureExitHandling(logger: Logger): void {
   process.on('uncaughtException', (err) => {
     logger.fatal({ err }, 'uncaughtException');
 
-    process.exit(1);
+    process.exitCode = 1;
```
gnarea, line 7: This approach is recommended over the previous one: https://nodejs.org/api/process.html#event-uncaughtexception

annaavanesyan (approved):

PR #20, opened 2023-02-22
chore: Fix ESLint violations in `env-var` test utils

gnarea:


gnarea:
Merged into #17 so it can all be reviewed/merged in one go.

PR #38, opened 2023-03-09
feat: Implement logic to create orgs

levansuper:
 

# Other information


src/services/routes/org.routes.spec.ts
```diff
@@ -16,20 +28,29 @@ describe('org routes', () => {
       url: '/orgs',
     };
 
-    test('Valid parameters with non ASCII name should return success', async () => {
+    test('Valid parameters should return urls', async () => {
```
gnarea, line 31:
```suggestion
    test('Valid parameters should return URLs', async () => {
```

Just for consistency with other repos.

src/services/routes/org.routes.ts
```diff
@@ -1,16 +1,28 @@
...
 import type { FastifyTypedInstance } from '../fastify.js';
-
-export enum ProblemType {
-  MALFORMED_ORG_NAME = 'https://veraid.net/problems/malformed-org-name',
-  MALFORMED_AWALA_ENDPOINT = 'https://veraid.net/problems/malformed-awala-endpoint',
+import { createOrg } from '../../org.js';
+import { CreationProblemType } from '../../CreationProblemType.js';
+
+const ORG_ROUTES_ERROR_MAPPING: {
+  [key in CreationProblemType]: (typeof HTTP_STATUS_CODES)[keyof typeof HTTP_STATUS_CODES];
```
gnarea, line 11: I'm sure you tried this, but just to double check: isn't there a simpler alternative to `(typeof HTTP_STATUS_CODES)[keyof typeof HTTP_STATUS_CODES]`?

Doesn't `keyof typeof HTTP_STATUS_CODES` work?

src/services/routes/org.routes.ts
```diff
@@ -1,16 +1,28 @@
...
 import { ORG_SCHEMA } from '../schema/org.schema.js';
 import type { FastifyTypedInstance } from '../fastify.js';
-
-export enum ProblemType {
-  MALFORMED_ORG_NAME = 'https://veraid.net/problems/malformed-org-name',
-  MALFORMED_AWALA_ENDPOINT = 'https://veraid.net/problems/malformed-awala-endpoint',
+import { createOrg } from '../../org.js';
+import { CreationProblemType } from '../../CreationProblemType.js';
+
+const ORG_ROUTES_ERROR_MAPPING: {
```
gnarea, line 10: Since this isn't exported, I'd suggest something along the lines of `RESPONSE_CODE_BY_PROBLEM_TYPE` or `RESPONSE_CODE_BY_PROBLEM`.

src/services/routes/org.routes.ts
```diff
@@ -1,16 +1,28 @@
...
+  [CreationProblemType.EXISTING_ORG_NAME]: HTTP_STATUS_CODES.CONFLICT,
+  [CreationProblemType.MALFORMED_AWALA_ENDPOINT]: HTTP_STATUS_CODES.BAD_REQUEST,
+  [CreationProblemType.MALFORMED_ORG_NAME]: HTTP_STATUS_CODES.BAD_REQUEST,
+} as const;
+
+interface OrgUrls {
+  self: string;
 }
 
+const formUrls = (name: string): OrgUrls => ({
```
gnarea, line 22: I'd say "form" is correct but it could potentially be ambiguous. How about `makeUrls()`?

src/services/routes/org.routes.ts
```diff
@@ -25,27 +37,24 @@ export default function registerRoutes(
...
-          type: ProblemType.MALFORMED_AWALA_ENDPOINT,
-        });
+      const result = await createOrg(request.body, {
+        logger: this.log,
+        dbConnection: this.mongoose,
+      });
+      if (result.didSucceed) {
+        await reply
+          .code(HTTP_STATUS_CODES.OK)
+          .header('Content-Type', 'application/json')
```
gnarea, line 47: Can you try dropping this? It shouldn't be needed -- plus, Fastify would automatically specify the charset, which is a good idea.

src/services/routes/org.routes.ts
```diff
@@ -25,27 +37,24 @@ export default function registerRoutes(
...
+          .send(formUrls(result.result.name));
         return;
       }
 
       await reply
-        .code(HTTP_STATUS_CODES.OK)
+        .code(ORG_ROUTES_ERROR_MAPPING[result.reason])
         .header('Content-Type', 'application/json')
```
gnarea, line 54: Same here.

gnarea (changes requested):

gnarea:
FYI @levansuper, I just changed the title to make it more user-friendly, since this will go on the release notes.

src/services/routes/org.routes.ts
levansuper: Yep I tried that.
 keyof typeof HTTP_STATUS_CODES returns the keys of the object like "OK", "NO_CONTENT", etc...
 In this case we need the values not the keys that's why I had to write this

gnarea: Cool, let's leave it as it is then 👍🏾

src/services/routes/org.routes.ts
levansuper: OK is very specific for the success case. we will not have reason in this case

gnarea: Note I'm referring to the `Content-Type`... I think that if we drop that line, Fastify will use `application/json; charset=utf-8`, which is even better.

gnarea (dismissed):
LGTM but CI is still blocked until you fix the merge conflicts with `main`.

gnarea (approved):
Awesome, thanks! 🚀

PR #44, opened 2023-03-16
feat: Implement org patch route

levansuper:


src/org.ts
```diff
@@ -3,38 +3,46 @@ import isValidDomain from 'is-valid-domain';
...
   OPEN: MemberAccessType.OPEN,
+} as const;
+
+type ReversedMemberAccessType = {
+  [key in MemberAccessType]: OrgSchema['memberAccessType'];
+};
+
+const REVERSE_MEMBER_ACCESS_MAPPING: ReversedMemberAccessType = {
+  inviteOnly: 'INVITE_ONLY',
+  open: 'OPEN',
```
levansuper, line 24: we might write a better mapping in the mapper functions if we decide to go with them

src/org.ts
```diff
@@ -73,3 +82,79 @@ export async function createOrg(
...
+    return {
+      didSucceed: false,
+      reason: OrgProblemType.ORG_NOT_FOUND,
+    };
+  }
+
+  return {
+    didSucceed: true,
+
+    // TODO : implement a mapper that transforms an OrgSchema to the OrgSchemaType object
```
levansuper, line 153: mapper functions might be useful tools. example:
transform(OrgSchema) -> OrgSchemaType
transform2(OrgSchemaType) -> OrgSchema

src/org.ts
```diff
@@ -73,3 +82,79 @@ export async function createOrg(
...
+    throw err as Error;
+  }
+
+  options.logger.info({ name: orgData.name }, 'Org updated');
+  return {
+    didSucceed: true,
+  };
+}
+
+export async function getOrg(
```
levansuper, line 127: I could move this method and the corresponding tests to a separate PR

src/models/Org.model.ts
```diff
@@ -1,7 +1,7 @@
 import { prop } from '@typegoose/typegoose';
 
 export enum MemberAccessType {
-  INVITE_ONLY = 'invite_only',
+  INVITE_ONLY = 'inviteOnly',
```
gnarea, line 4: I think camelCase is fine for identifiers in source code, but looks really odd to use on an enum value or a DB field value.

src/services/routes/org.routes.spec.ts
```diff
@@ -186,7 +188,147 @@ describe('org routes', () => {
...
+
+      const response = await serverInstance.inject({
+        ...injectionOptions,
+        payload,
+      });
+
+      expect(response).toHaveProperty('statusCode', 200);
+    });
+
+    test('name matching the url parameter should be accepted', async () => {
```
gnarea, line 225: Nitpick: The convention is to title-case these strings, so "n" in "name" should be uppercase.

This also applies to other tests in this file.

src/services/routes/org.routes.spec.ts
```diff
@@ -186,7 +188,147 @@ describe('org routes', () => {
...
+      mockUpdateOrg.mockResolvedValueOnce({
+        didSucceed: true,
+      });
+
+      const response = await serverInstance.inject({
+        ...injectionOptions,
+        payload,
+      });
+
+      expect(response).toHaveProperty('statusCode', 200);
```
gnarea, line 240: Please use `HTTP_STATUS_CODES`. Same in the rest of the test file.

src/services/routes/org.routes.spec.ts
```diff
@@ -186,7 +188,147 @@ describe('org routes', () => {
...
+    });
+
+    test('non matching name should resolve into bad request status', async () => {
```
gnarea, line 297: I think this test should do two things:

- Check that the response is a 40X with the respective problem type (as it does  right now).
- Check that we haven't called `updateOrg()`.

Consequently, the title would then become something like "Non-matching name should be refused".

src/services/routes/org.routes.ts
```diff
@@ -22,7 +36,6 @@ interface OrgUrls {
 const makeUrls = (name: string): OrgUrls => ({
   self: `/orgs/${name}`,
 });
-
```
gnarea, line 25: Nitpick: we probably should keep this empty line for consistency.

ESLint allows both approaches because `makeUrls` is a variable, and variables are allowed to precede a function in the even the function uses that variable. However, in this particular case, `makeUrls` shouldn't be a variable, but a literal function... but that's a separate conversation as that's outside the scope of this PR.

src/services/routes/org.routes.ts
```diff
@@ -52,5 +65,41 @@ export default function registerRoutes(
...
+      if (!getOrgResult.didSucceed) {
+        await reply.code(RESPONSE_CODE_BY_PROBLEM[getOrgResult.reason]).send({
+          type: getOrgResult.reason,
+        });
+        return;
+      }
+
+      const result = await updateOrg(orgName, request.body, serviceOptions);
+      if (result.didSucceed) {
+        await reply.code(HTTP_STATUS_CODES.OK).send();
```
gnarea, line 94: I think this is ambiguous from a code perspective, and _maybe_ invalid as well. What's the response here? `text/plain` with an empty body? `application/json` with an empty body? The latter would be invalid JSON.

I think returning nothing is perfectly acceptable, but in that case we should return a `204` response.

src/utilities/result.ts
```diff
@@ -7,9 +7,8 @@ export interface FailureResult<Reason> extends BaseResult {
   readonly reason: Reason;
 }
 
-export interface SuccessfulResult<Result> extends BaseResult {
-  readonly didSucceed: true;
-  readonly result: Result;
-}
+export type SuccessfulResult<Result> = Result extends undefined
```
gnarea, line 10: Is this change really necessary? Couldn't we just do `Result<void, FailureReason>`?

If we did need this change, or a version of it, would we really need to do `BaseResult & ...` in the lines below? It seems redundant.

src/org.spec.ts
```diff
@@ -255,4 +255,365 @@ describe('org', () => {
       expect(error).toHaveProperty('name', 'MongoNotConnectedError');
     });
   });
+
+  describe('updateOrg', () => {
+    test('Valid data should be allowed and updated', async () => {
```
gnarea, line 260: This title suggests we're going to check for two things, but we only check for one (whether the DB was updated). We're not checking whether the result is successful.

However, I think we should do those checks in separate tests, just like we did in the creation test suite above.

src/org.spec.ts
```diff
@@ -255,4 +255,365 @@ describe('org', () => {
...
+          dbConnection: connection,
+          logger: mockLogging.logger,
+        },
+      );
+
+      requireFailureResult(result);
+      expect(result.reason).toBe(OrgProblemType.MALFORMED_ORG_NAME);
+    });
+
+    test('Non matching name should be refused', async () => {
```
gnarea, line 507: Related tests together.

src/org.spec.ts
```diff
@@ -255,4 +255,365 @@ describe('org', () => {
...
+        awalaEndpoint: nonAsciiAwalaEndpoint,
+      });
+      expect(dbResult).not.toBeNull();
+    });
+
+    test('Malformed name should be refused', async () => {
+      const connection = getConnection();
+
+      const result = await updateOrg(
+        'INVALID_Name',
```
gnarea, line 493: I'd factor out  `'INVALID_Name` to a variable just to be explicit about our intent for the two to match.

src/org.spec.ts
```diff
@@ -255,4 +255,365 @@ describe('org', () => {
...
+            logger: mockLogging.logger,
+          }),
+        Error,
+      );
+
+      expect(error).toHaveProperty('name', 'MongoNotConnectedError');
+    });
+
+    describe('getOrg', () => {
+      test('valid name should return the corresponding data', async () => {
```
gnarea, line 566: It's not so much that the name is _valid_ what's important, it's that it _exists_.

src/org.spec.ts
```diff
@@ -255,4 +255,365 @@ describe('org', () => {
...
+
+        requireSuccessfulResult(methodResponse);
+        expect(methodResponse.result).toMatchObject({
+          name: ORG_NAME,
+          memberAccessType: 'OPEN',
+          awalaEndpoint: AWALA_ENDPOINT,
+        });
+      });
+
+      test('invalid name should return NOT_FOUND error', async () => {
```
gnarea, line 590: Similarly, I'd say "invalid" -> "non-existing".

Also, `NOT_FOUND` suggests there's an identifier by that name, probably one for the `404` response. I'd avoid that... maybe something like `Non-existing name should be refused`?

src/org.ts
gnarea: Let's just keep it in mind for the future, since for the purpose of this PR, as simple existance check would've sufficed. But since it's already here, let's keep it.

src/org.ts
```diff
@@ -73,3 +82,79 @@ export async function createOrg(
...
+  });
+  let org;
+
+  try {
+    org = await orgModel.findOne({
+      name,
+    });
+  } catch (err) {
+    throw err as Error;
+  }
```
gnarea, line 142: Why catch the error then?

src/org.ts
gnarea: Agreed

gnarea (changes requested):

gnarea:
Only overarching comment I have is that we should consider factoring out / parametising tests to avoid duplication -- for example, when doing schema validation in the creation and update tests.

src/services/routes/org.routes.ts
levansuper: change to function

src/org.ts
gnarea: If we're not implementing that in this PR, which is totally fine, then let's please create an issue and delete this TODO. Otherwise, we'll just forget about it -- also, TODOs introduce a ESLint warnings.

gnarea (changes requested):
Thanks!

Nearly there! I just added a new comment, and left one unresolved as I think it hasn't been addressed yet.

gnarea (approved):
Awesome, thanks! 🚀

PR #54, opened 2023-03-21
test: Update create org tests

levansuper:
 

# Other information

### DIFF
--- a/src/services/routes/org.routes.spec.ts
+++ b/src/services/routes/org.routes.spec.ts
@@ -3,7 +3,12 @@ import type { InjectOptions } from 'fastify';
 import { jest } from '@jest/globals';
 
 import { configureMockEnvVars, REQUIRED_SERVER_ENV_VARS } from '../../testUtils/envVars.js';
-import { AWALA_ENDPOINT, NON_ASCII_AWALA_ENDPOINT, ORG_NAME } from '../../testUtils/stubs.js';
+import {
+  AWALA_ENDPOINT,
+  NON_ASCII_AWALA_ENDPOINT,
+  NON_ASCII_ORG_NAME,
+  ORG_NAME,
+} from '../../testUtils/stubs.js';
 import {
   type OrgSchema,
   type OrgSchemaPatch,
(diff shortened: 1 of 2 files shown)


src/testUtils/server.ts
```diff
@@ -0,0 +1,22 @@
...
+    server = await makeServer();
+  });
+
+  beforeEach(async () => {
+    await server.close();
+    server = await makeServer();
+  });
+
+  afterEach(async () => {
+    await server.close();
```
gnarea, line 18: Have you checked what happens when `server.close()` is called multiple times?

I'm asking because, under certain circumstances, Fastify would internally call `server.close()` when you use `server.inject()` (like we do in our tests). For example, in the event of internal server errors.

gnarea: To clarify, it's a genuine question -- I don't know what happens when you call it twice. I'm just saying we should check what happens because we might have to call `server.close()` conditionally if it can only ever be called once.

src/testUtils/server.ts
```diff
@@ -0,0 +1,22 @@
...
+  });
```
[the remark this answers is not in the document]
levansuper: This is a nice catch. Changing that

src/testUtils/server.ts
levansuper: This does not seam to have any issue. I tested it and called close 2 times without any error

gnarea: Fantastic, let's ignore this then 🙌🏾

src/services/routes/org.routes.spec.ts
```diff
@@ -37,7 +37,7 @@ describe('org routes', () => {
   configureMockEnvVars(REQUIRED_SERVER_ENV_VARS);
   const getTestServer = setUpTestServer();
   let serverInstance: FastifyTypedInstance;
-  beforeEach(() => {
+  beforeAll(() => {
```
gnarea, line 40: The original `beforeEach()` was correct. This must be working by chance -- Note that you're now caching the instance across tests.

src/testUtils/server.ts
```diff
@@ -9,12 +7,7 @@ export function setUpTestServer(): () => FastifyTypedInstance {
     server = await makeServer();
```
gnarea, line 7: You forgot to change the `beforeAll()` above to a `beforeEach()`.

gnarea (changes requested):
Note that your latest changes are now caching the same server instance and reusing it across tests. See in-line comments.

gnarea (approved):
Thanks! 🚀

PR #58, opened 2023-03-23
feat: Implement org deletion endpoint

### LINKED ISSUE #5: Create CRUD endpoints for organisations
`POST /orgs`, `GET /orgs/{org}/`, `PATCH /orgs/{org}/` and `DELETE /orgs/{org}/`, as described on the README in principle, but as always feedback is most welcomed.

Probably best to create one PR per endpoint.

Skip auth enforcement if #2 isn't implemented.

levansuper:
Fixes #5.

### DIFF
--- a/src/org.spec.ts
+++ b/src/org.spec.ts
@@ -1,7 +1,6 @@
 /* eslint-disable unicorn/text-encoding-identifier-case */
 import { getModelForClass, type ReturnModelType } from '@typegoose/typegoose';
 import type { Connection } from 'mongoose';
-import { jest } from '@jest/globals';
 
 import { MemberAccessType, OrgModelSchema } from './models/Org.model.js';
 import { createOrg, deleteOrg, getOrg, updateOrg } from './org.js';
@@ -518,17 +517,6 @@ describe('org', () => {
         name: NON_ASCII_ORG_NAME,
       });
       expect(dbResult).not.toBeNull();
-    });
-
-    test('Non existing name should not produce logs', async () => {
-      const info = jest.spyOn(mockLogging.logger, 'info');
-
-      await deleteOrg(ORG_NAME, {
-        dbConnection: connection,
-        logger: mockLogging.logger,
-      });
-
-      expect(info).not.toHaveBeenCalled();
     });
 
     test('Record deletion errors should be propagated', async () => {
--- a/src/org.spec.ts
+++ b/src/org.spec.ts
@@ -522,6 +522,7 @@ describe('org', () => {
 
     test('Non existing name should not produce logs', async () => {
       const info = jest.spyOn(mockLogging.logger, 'info');
+
       await deleteOrg(ORG_NAME, {
         dbConnection: connection,
         logger: mockLogging.logger,
--- a/src/org.spec.ts
+++ b/src/org.spec.ts
@@ -3,7 +3,7 @@ import { getModelForClass, type ReturnModelType } from '@typegoose/typegoose';
 import type { Connection } from 'mongoose';
 
 import { MemberAccessType, OrgModelSchema } from './models/Org.model.js';
-import { createOrg, getOrg, updateOrg } from './org.js';
+import { createOrg, deleteOrg, getOrg, updateOrg } from './org.js';
 import {
   type OrgSchema,
   type OrgSchemaMemberAccessType,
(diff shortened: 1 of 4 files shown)


src/services/routes/org.routes.spec.ts
```diff
@@ -331,7 +333,7 @@ describe('org routes', () => {
     });
   });
 
-  describe('get/:orgName', () => {
+  describe('get by name', () => {
```
levansuper, line 336: Thought this get/:orgName is not the best name

src/org.spec.ts
```diff
@@ -478,4 +478,60 @@ describe('org', () => {
...
+      });
+
+      requireSuccessfulResult(methodResponse);
+      const dbResult = await orgModel.exists({
+        name: name,
+      });
+      expect(dbResult).toBeNull();
+    });
+
+    test('Non existing name should not remove any org', async () => {
```
levansuper, line 504: This test makes sure we don't accidentally write something line orgModel.delete({})
Maybe we can come up with a better name

src/org.spec.ts
```diff
@@ -478,4 +478,60 @@ describe('org', () => {
...
+      });
+
+      requireSuccessfulResult(methodResponse);
+      const dbResult = await orgModel.exists({
+        name: NON_ASCII_ORG_NAME,
+      });
+      expect(dbResult).not.toBeNull();
+    });
+
+    test('Record Find errors should be propagated', async () => {
```
gnarea, line 522:
```suggestion
    test('Record deletion errors should be propagated', async () => {
```

src/services/routes/org.routes.ts
```diff
@@ -131,5 +131,26 @@ export default function registerRoutes(
...
 
+  fastify.route({
+    method: ['DELETE'],
+    url: '/orgs/:orgName',
+
+    schema: {
+      params: ORG_ROUTE_PARAMS,
+    },
+
+    async handler(request, reply): Promise<void> {
```
gnarea, line 142: I think we should return a 404 when the org isn't found. It's what a RESTFul API is supposed to do and what we've done in other endpoints. If `PATCH /orgs/example.com` results in a 404, why does `DELETE /orgs/example.com` not?

src/org.ts
```diff
@@ -140,3 +140,19 @@ export async function getOrg(
...
+  options: ServiceOptions,
+): Promise<Result<undefined, OrgProblemType>> {
+  const orgModel = getModelForClass(OrgModelSchema, {
+    existingConnection: options.dbConnection,
+  });
+  await orgModel.deleteOne({
+    name,
+  });
+
+  return {
```
gnarea, line 155: Log, please. Also make sure to update the existing happy-path test.

gnarea (changes requested):
Thanks! Nearly there

src/org.spec.ts
```diff
@@ -507,19 +508,29 @@ describe('org', () => {
...
-      requireSuccessfulResult(methodResponse);
+      requireSuccessfulResult(result);
       const dbResult = await orgModel.exists({
         name: NON_ASCII_ORG_NAME,
       });
       expect(dbResult).not.toBeNull();
     });
 
-    test('Record Find errors should be propagated', async () => {
+    test('Non existing name should not produce logs', async () => {
```
gnarea, line 523: Keep in mind that the convention is not to create a test just to check logs, but to check the log in the appropriate test.

src/org.ts
levansuper: I thought this would be a better way of logging. To check if it actually deleted something. If yes then I think we could update the "updateOrg" method to check if it really updated something too

src/org.ts
```diff
@@ -148,10 +148,14 @@ export async function deleteOrg(
...
     existingConnection: options.dbConnection,
   });
-  await orgModel.deleteOne({
+  const result = await orgModel.deleteOne({
     name,
   });
 
+  if (result.deletedCount > 0) {
+    options.logger.info({ name }, 'Org deleted');
+  }
```
gnarea, line 157: Let's keep things simple and unconditionally log that message.

This `if` is only guarding against a very edge case, where two simultaneous `DELETE /orgs/:orgName` come in for the same org... Which isn't worth the hassle.

src/services/routes/org.routes.spec.ts
```diff
@@ -392,24 +392,53 @@ describe('org routes', () => {
...
         ...injectionOptions,
-        url: `/orgs/${name}`,
+        url: `/orgs/${ORG_NAME}`,
       });
 
-      expect(mockDeleteOrg).toHaveBeenCalledWith(name, {
+      expect(mockGetOrg).toHaveBeenCalledWith(ORG_NAME, {
+        logger: serverInstance.log,
+        dbConnection: serverInstance.mongoose,
+      });
```
gnarea, line 416: I don't think it matters that we used `mockGetOrg` -- what matters is the effect (i.e., 404 response if non-existing), which is already checked in the other test.

The distinction will matter once we eventually replace that `getOrg()` call with a Fastify validator/plugin that checks the existence and auth.

src/services/routes/org.routes.spec.ts
```diff
@@ -392,24 +392,53 @@ describe('org routes', () => {
...
+
+      const response = await serverInstance.inject({
+        ...injectionOptions,
+        url: `/orgs/${ORG_NAME}`,
+      });
+
+      expect(mockGetOrg).toHaveBeenCalledWith(ORG_NAME, {
+        logger: serverInstance.log,
+        dbConnection: serverInstance.mongoose,
+      });
```
gnarea, line 438: Again, I don't think it matters we used `mockGetOrg`, especially because we're seeing its effect in the 404 response.

gnarea (changes requested):

gnarea (approved):
Awesome! 🚀

PR #66, opened 2023-03-29
feat: Implement get member route

levansuper:
 

# Other information

### DIFF
--- a/src/member.spec.ts
+++ b/src/member.spec.ts
@@ -4,7 +4,13 @@ import type { Connection } from 'mongoose';
 
 import { setUpTestDbConnection } from './testUtils/db.js';
 import { makeMockLogging, type MockLogging, partialPinoLog } from './testUtils/logging.js';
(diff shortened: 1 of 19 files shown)


src/member.spec.ts
```diff
@@ -31,6 +31,7 @@ describe('member', () => {
   let connection: Connection;
   let serviceOptions: ServiceOptions;
   let memberModel: ReturnModelType<typeof MemberModelSchema>;
+  const mongoId = '6424ad273f75645b35f9ee79';
```
gnarea, line 34: I think it's confusing to define `mongoId` here, when it's actually only needed in the non-existing case. 

`Record Find errors should be propagated` creates an actual record, so we could/should just use that one.

gnarea (changes requested):
Thanks! Only a couple of minor things.

gnarea (approved):
Thanks!

PR #69, opened 2023-03-31
feat: Implement delete member

levansuper:


src/services/routes/member.routes.ts
```diff
@@ -115,5 +115,34 @@ export default function registerRoutes(
...
+
+      const getMemberResult = await getMember(orgName, memberId, serviceOptions);
+      if (!getMemberResult.didSucceed) {
+        await reply.code(RESPONSE_CODE_BY_PROBLEM[getMemberResult.reason]).send({
+          type: getMemberResult.reason,
+        });
+        return;
+      }
+
+      await deleteMember(orgName, serviceOptions);
```
gnarea, line 141:
```suggestion
      await deleteMember(memberId, serviceOptions);
```

src/testUtils/stubs.ts
```diff
@@ -5,3 +5,4 @@ export const NON_ASCII_AWALA_ENDPOINT = 'はじめよう.みんな';
 export const MEMBER_NAME = 'TEST_MEMBER_NAME';
 export const NON_ASCII_MEMBER_NAME = 'テスト ユーザー';
 export const MEMBER_EMAIL = 'membet@example.com';
+export const MEMBER_ID = '6424ad273f75645b35f9ee79';
```
gnarea, line 8: I'd call this something like `MEMBER_API_ID`. "Member id" is a different thing in VeraId.

gnarea (changes requested):
Thanks!

src/testUtils/stubs.ts
levansuper: Maybe MEMBER_MONGO_ID?

gnarea: Agreed. That's more accurate.

gnarea (approved):

src/testUtils/stubs.ts
gnarea: I just approved the PR just in case, but feel free to use that name and I'll re-approve it.

gnarea (approved):
Thanks!

PR #70, opened 2023-04-01
test: Fix flaky test in org creation

### LINKED ISSUE #68: Flaky test breaks CI ocassionally
https://github.com/relaycorp/veraid-authority/actions/runs/4564325901/jobs/8053948900

```
  ● org › createOrg › Clash with existing name should be refused

    expect(received).toBe(expected) // Object.is equality

    Expected: false
    Received: true

      10 |   result: Result<undefined, FailureReason>,
      11 | ): asserts result is FailureResult<FailureReason> {
    > 12 |   expect(result.didSucceed).toBe(false);
         |                             ^
      13 | }
      14 |

      at requireFailureResult (src/testUtils/result.ts:12:29)
      at Object.<anonymous> (src/org.spec.ts:102:7)
````

### PR DESCRIPTION
**Pull request by gnarea**
Fixes #68.

### DIFF
--- a/src/org.spec.ts
+++ b/src/org.spec.ts
@@ -96,6 +96,7 @@ describe('org', () => {
       };
       await createOrg(orgData, serviceOptions);
 
+      // Use different data to make sure the creation operation isn't being deduped
       const methodResult = await createOrg(
         { ...orgData, memberAccessType: 'OPEN' },
         serviceOptions,
--- a/src/org.spec.ts
+++ b/src/org.spec.ts
@@ -94,10 +94,12 @@ describe('org', () => {
         name: ORG_NAME,
         memberAccessType: 'INVITE_ONLY',
       };
-
       await createOrg(orgData, serviceOptions);
 
-      const methodResult = await createOrg(orgData, serviceOptions);
+      const methodResult = await createOrg(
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/org.spec.ts [hunk 1]
```diff
@@ -94,10 +94,13 @@ describe('org', () => {
         name: ORG_NAME,
         memberAccessType: 'INVITE_ONLY',
       };
-
       await createOrg(orgData, serviceOptions);
 
-      const methodResult = await createOrg(orgData, serviceOptions);
+      // Use different data to make sure the creation operation isn't being deduped
```
levansuper, line 99: do we need this comment?

levansuper: And should make the same change in members?

gnarea: Without the comment, I fear this is bound to happen: It's going to look really odd to use slightly different data for apparently no reason, so we're going to undo this PR because it'll still work on our laptops... But then, it'll fail on CI like it does here. But only sometimes.

It wasn't obvious to me to figure out what was going on, so I think we should avoid this in the future.

---

> And should make the same change in members?

Good question -- I forgot about that. I wonder why we haven't seen that other test fail.

I think we should, for good measure. I'll do it on Monday morning as I won't be able to use my computer again until then.

levansuper (approved):

PR #71, opened 2023-04-01
feat: Implement member update

### LINKED ISSUE #6: Create CRUD endpoints for org members
As described on the README in principle, but as always feedback is most welcomed.

Probably best to create one PR per endpoint.

Skip auth enforcement if #2 isn't implemented.

levansuper:
Fixes #6

### DIFF
--- a/src/member.spec.ts
+++ b/src/member.spec.ts
@@ -417,7 +417,7 @@ describe('member', () => {
       await connection.close();
 
       const error = await getPromiseRejection(
-        async () => updateMember(MEMBER_API_ID, {}, serviceOptions),
+        async () => updateMember(MEMBER_MONGO_ID, {}, serviceOptions),
         Error,
       );
       expect(error).toHaveProperty('name', 'MongoNotConnectedError');
--- a/src/member.spec.ts
+++ b/src/member.spec.ts
@@ -4,7 +4,13 @@ import type { Connection } from 'mongoose';
 
 import { setUpTestDbConnection } from './testUtils/db.js';
 import { makeMockLogging, type MockLogging, partialPinoLog } from './testUtils/logging.js';
-import { MEMBER_EMAIL, MEMBER_NAME, NON_ASCII_MEMBER_NAME, ORG_NAME } from './testUtils/stubs.js';
+import {
+  MEMBER_EMAIL,
+  MEMBER_NAME,
+  NON_ASCII_MEMBER_NAME,
+  NON_ASCII_ORG_NAME,
+  ORG_NAME,
+} from './testUtils/stubs.js';
 import type { ServiceOptions } from './serviceTypes.js';
 import { MemberModelSchema, Role } from './models/Member.model.js';
 import { createMember, getMember } from './member.js';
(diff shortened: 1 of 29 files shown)


src/member.ts
```diff
@@ -102,3 +102,38 @@ export async function deleteMember(
...
+  }
+
+  const role = memberData.role && ROLE_MAPPING[memberData.role];
+
+  const memberModel = getModelForClass(MemberModelSchema, {
+    existingConnection: options.dbConnection,
+  });
+
+  try {
+    await memberModel.findByIdAndUpdate(memberId, { ...memberData, role });
```
gnarea, line 123: Does this work when `memberData.role` isn't set?

src/member.ts
```diff
@@ -102,3 +102,38 @@ export async function deleteMember(
...
+  memberId: string,
+  memberData: PatchMemberSchema,
+  options: ServiceOptions,
+): Promise<Result<undefined, MemberProblemType>> {
+  const validationFailure = validateMemberData(memberData, options);
+  if (validationFailure !== undefined) {
+    return { didSucceed: false, reason: validationFailure };
+  }
+
+  const role = memberData.role && ROLE_MAPPING[memberData.role];
```
gnarea, line 116:
```suggestion
  const role = memberData.role ?? ROLE_MAPPING[memberData.role];
```

That way, `role` type becomes `MemberRole | null`. (If I'm not mistaken, right now it's `MemberRole | false`)

src/member.spec.ts
```diff
@@ -261,4 +261,166 @@ describe('member', () => {
       expect(error).toHaveProperty('name', 'MongoNotConnectedError');
     });
   });
+
+  describe('updateMember', () => {
+    const testEmail = 'different@email.com';
+
+    test('Valid should be updated', async () => {
```
gnarea, line 268: This is testing the scenario where all the fields are set, but I think the opposite (nothing is changed) is also useful.

src/services/routes/member.routes.spec.ts
```diff
@@ -273,4 +277,135 @@ describe('member routes', () => {
...
+
+      const response = await serverInstance.inject({
+        ...injectionOptions,
+        payload: {},
+      });
+
+      expect(response).toHaveProperty('statusCode', HTTP_STATUS_CODES.NO_CONTENT);
+    });
+
+    test.each(memberSchemaRoles)('Role %s should be allowed', async (role: MemberSchemaRole) => {
```
gnarea, line 311: I'm not sure this test is useful.

gnarea (changes requested):

src/services/routes/member.routes.spec.ts
levansuper: This tests the role enum in the schema

gnarea (changes requested):

src/models/Member.model.ts
```diff
@@ -7,13 +7,17 @@ export enum Role {
...
   { orgName: 1, name: 1 },
-  { unique: true, partialFilterExpression: { name: { $exists: true } } },
+  { unique: true, partialFilterExpression: { name: { $type: 'string' } } },
 )
 export class MemberModelSchema {
-  @prop()
+  @prop({
+    default: null,
+  })
   public name?: string;
```
gnarea, line 16: This type is `string | undefined`, which is wrong. It should be `string | null`:

```suggestion
  public name!: string | null;
```

gnarea (changes requested):

gnarea (approved):
Thanks!

PR #76, opened 2023-04-04
fix: Implement KMS-backed key pair for each organisation

### LINKED ISSUE #4: Implement KMS-backed key pair for each organisation
Some code in `@relaycorp/awala-keystore-cloud` would be a good starting point.

gnarea:
This PR integrates [`@relaycorp/webcrypto-kms`](https://github.com/relaycorp/webcrypto-kms-js) in such a way that we use a [mock AWS KMS server](https://github.com/nsmithuk/local-kms) in development and can use any supported KMS in production (GCP's in our case).

Fixes #4.

# Manual testing

I tested it with the following HTTP file in WebStorm:

```http
### Create org
POST http://veraid-authority.default.10.103.177.106.sslip.io/orgs
Content-Type: application/json

{
  "name": "example.com",
  "memberAccessType": "OPEN"
}

### Destroy org

DELETE http://veraid-authority.default.10.103.177.106.sslip.io/orgs/example.com
```

And the Local KMS logs showed the key was created and subsequently removed:

```
INFO[2023-04-06 22:36:17.482] Local KMS started on 0.0.0.0:8080            
INFO[2023-04-06 22:42:02.194] New RSA_2048 key created: arn:aws:kms:eu-west-2:111122223333:key/88d88fac-abe7-49c4-8f83-a480d2facf8a 
INFO[2023-04-06 22:43:49.804] Schedule key deletion: arn:aws:kms:eu-west-2:111122223333:key/88d88fac-abe7-49c4-8f83-a480d2facf8a 
```

### DIFF
--- a/.idea/runConfigurations/_template__of_JavaScriptTestRunnerJest.xml
+++ b/.idea/runConfigurations/_template__of_JavaScriptTestRunnerJest.xml
@@ -1,0 +1,10 @@
+<!-- Work around for https://youtrack.jetbrains.com/issue/WEB-52967/Automatically-add-experimental-vm-modules-to-Jest-for-ESM-projects -->
+<component name="ProjectRunConfigurationManager">
+  <configuration default="true" type="JavaScriptTestRunnerJest">
+    <node-interpreter value="project" />
+    <node-options value="--experimental-vm-modules" />
+    <envs />
+    <scope-kind value="ALL" />
+    <method v="2" />
+  </configuration>
+</component>
--- a/.idea/runConfigurations/_template__of_JavaScriptTestRunnerJest.xml
+++ b/.idea/runConfigurations/_template__of_JavaScriptTestRunnerJest.xml
@@ -1,4 +1,3 @@
-<!-- Work around for https://youtrack.jetbrains.com/issue/WEB-52967/Automatically-add-experimental-vm-modules-to-Jest-for-ESM-projects -->
 <component name="ProjectRunConfigurationManager">
   <configuration default="true" type="JavaScriptTestRunnerJest">
     <node-interpreter value="project" />
--- a/src/org.ts
+++ b/src/org.ts
@@ -159,11 +159,19 @@ export async function deleteOrg(
   const orgModel = getModelForClass(OrgModelSchema, {
     existingConnection: options.dbConnection,
   });
+  const org = await orgModel.findOne({ name });
 
-  await orgModel.deleteOne({
-    name,
-  });
-  options.logger.info({ name }, 'Org deleted');
+  if (org === null) {
+    options.logger.info({ name }, 'Ignored deletion of non-existing org');
+  } else {
+    const kms = await Kms.init();
+    const privateKey = await kms.retrievePrivateKeyByRef(org.privateKeyRef);
+    await kms.destroyPrivateKey(privateKey);
+
+    await org.deleteOne();
+
+    options.logger.info({ name }, 'Org deleted');
+  }
 
   return {
     didSucceed: true,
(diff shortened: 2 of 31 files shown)


.idea/runConfigurations/_template__of_JavaScriptTestRunnerJest.xml
```diff
@@ -0,0 +1,9 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="true" type="JavaScriptTestRunnerJest">
+    <node-interpreter value="project" />
+    <node-options value="--experimental-vm-modules" />
+    <envs />
+    <scope-kind value="ALL" />
+    <method v="2" />
+  </configuration>
+</component>
```
levansuper, line 9: should this be in this PR?

gnarea: No, it must've been added automatically when I merged the PR this one is based on.

BTW, this is a draft PR, so not ready to be reviewed yet.

gnarea: This is gone now that I've merged `main`.

levansuper (commented):
Question. Should there be a way to recreate the key pair?

gnarea:
> Should there be a way to recreate the key pair?

Do you mean: (a) rotate the key pair (e.g., if the private key is compromised), or (b) restore the key pair if someone accidentally delete the org?

(a) I'd be very keen to support something like that in the future. Not so much because of a key compromise, as that's extremely unlikely, but because we want to migrate to better algorithms in the future (from RSA to EdDSA). However, I'd prefer to do it with concrete requirements in mind -- otherwise we might be implementing something based on too many hypothesis.

(b) That'd depend on the underlying KMS. Both GCP and AWS give you "grace periods" of 1+ days, during which you can abort the destruction of the key. However, I'd rather leave that to any operator of this server as it could get very complicated.

src/testUtils/webcrypto.ts
```diff
@@ -0,0 +1,9 @@
+import { Crypto } from '@peculiar/webcrypto';
+
+import { KEY_USAGES, RSA_PSS_CREATION_ALGORITHM } from '../utilities/kms/keyParams.js';
+
+export const NODEJS_PROVIDER = new Crypto().subtle;
+
+export async function generateKeyPair(): Promise<CryptoKeyPair> {
+  return NODEJS_PROVIDER.generateKey(RSA_PSS_CREATION_ALGORITHM, true, KEY_USAGES);
+}
```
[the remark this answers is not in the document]
gnarea: I think so. I added a note there.

src/org.ts
```diff
@@ -68,7 +77,7 @@ export async function createOrg(
   options.logger.info({ name: orgData.name }, 'Org created');
   return {
     didSucceed: true,
-    result: { name: org.name },
+    result: { name: orgData.name },
```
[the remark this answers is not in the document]
gnarea: Why though? The DB shouldn't be transforming the org name in any way (if it did, that'd be a bug).

Note that line 77 was already using `orgData.name` instead of the original `org.name`, so part of the rationale for this change was consistency.

The main reason for making this change, though, is that the argument to `orgModel.create()` is now more complicated as it contains two new fields (`privateKeyRef` and `publicKey`) -- both of which need to be computed. I didn't want to do of all that complicated stuff between the `try`/`catch`, so I decided to assign it to another variable (`org`) outside the `try`/`catch`.

So if we stick to the original approach of reading the output of `orgModel.create(...)`, we're going to end up with two instances of the same org (`inputOrg` and `outputOrg`, for example)... Even though we don't really need two instances of the same org.

levansuper (approved):

PR #77, opened 2023-04-04
test: Make tests work locally again

gnarea:
I hadn't run the tests in the past week, and went to run them today but it didn't work for two very different reasons:

- Some **local** modules could no longer be resolved, getting errors like https://github.com/facebook/jest/issues/12284. This failed with both `npm test`  and WebStorm's selective test runs. It looks like `package-lock.json` was corrupted somehow (bad merge?), so I fixed it by re-creating the lock file as I couldn't locate the offending entry(ies).
- [The latest WebStorm no longer adds `--experimental-vm-modules`](https://youtrack.jetbrains.com/issue/WEB-52967/Automatically-add-experimental-vm-modules-to-Jest-for-ESM-projects#focus=Comments-27-6784178.0-0), so I'm adding a configuration template to make sure the flag is added.

### DIFF
--- a/.idea/runConfigurations/_template__of_JavaScriptTestRunnerJest.xml
+++ b/.idea/runConfigurations/_template__of_JavaScriptTestRunnerJest.xml
@@ -1,0 +1,10 @@
+<!-- Work around for https://youtrack.jetbrains.com/issue/WEB-52967/Automatically-add-experimental-vm-modules-to-Jest-for-ESM-projects -->
+<component name="ProjectRunConfigurationManager">
+  <configuration default="true" type="JavaScriptTestRunnerJest">
+    <node-interpreter value="project" />
+    <node-options value="--experimental-vm-modules" />
+    <envs />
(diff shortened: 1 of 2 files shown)


levansuper (approved):
Reinstalled packages and ran tests.
Looks to be working!

levansuper:
Not sure why but had a problem with this
What i did was fetch the branch, `npm install`, run tests and git status. `.idea/runConfigurations/_template__of_JavaScriptTestRunnerJest.xml` was changed

gnarea:
Changed how? Can you show me a diff or screenshot?

I did get that with the first commit, which is why I changed the file in the second commit.

I wonder whether you're using WebStorm 2022.X and that version changes the file in a different way. If that's the case, I'd suggest upgrading to version 2023.X. Thus should be a temporary workaround anyway so the version of WS should not matter for too long.

levansuper:
Screenshot 2023-04-05 at 18 38 06

This happens when i run `npm install`

PR #78, opened 2023-04-05
feat: Implement member public key create and delete routes

### LINKED ISSUE #7: Create member key creation and deletion endpoints
- Key creation time must be stored for auditing purposes, but the last update time should not be stored. See this example: https://github.com/relaycorp/awala-keystore-cloud-js/blob/6b1499e5f76d5d9081177743028cc25f6afb54b7/src/lib/gcp/models/GcpSessionKey.ts#L4 (note that `creationDate` is set explicitly on line 10)
- The public key MUST use the RSA algorithm with a modulus of at least 2048 bit.

---

Heads up: Here's how you can generate public keys to use in the tests (I haven't tested it):

```ts
import { generateKeyPair } from 'node:crypto';
import { promisify } from 'node:util';

const generateKeyPairAsync = promisify(generateKeyPair);

const { publicKey } = await generateKeyPairAsync('rsa-pss', {
  modulusLength: 2048,
  publicKeyEncoding: {type: 'spki', format: 'der'},
});
```

This is an expensive operation, so best to reuse the same valid key across tests.

Here's some examples of invalid keys (for the tests where we check invalid keys are refused):

```ts
// Invalid modulus
await generateKeyPairAsync('rsa-pss', {
  modulusLength: 2047,
  publicKeyEncoding: {type: 'spki', format: 'der'},
});

// Invalid algorithm
await generateKeyPairAsync('dsa', {
  modulusLength: 2048,
  publicKeyEncoding: {type: 'spki', format: 'der'},
});
```

levansuper:
fixes #7


src/services/routes/memberPublicKey.routes.ts
```diff
@@ -0,0 +1,111 @@
+import type { RouteOptions } from 'fastify';
+
+import { HTTP_STATUS_CODES } from '../http.js';
+import type { PluginDone } from '../types/PluginDone.js';
+import type { FastifyTypedInstance } from '../fastify.js';
+import { MemberPublicKeyProblemType } from '../../businessLogic/memberPublicKey/MemberPublicKeyProblemType.js';
```
levansuper, line 6: Maybe we need to rename things?

src/services/schema/memberPublicKey.schema.ts
```diff
@@ -0,0 +1,14 @@
+import type { FromSchema } from 'json-schema-to-ts';
+
+export const MEMBER_PUBLIC_KEY_SCHEMA = {
+  type: 'object',
+
+  properties: {
+    publicKey: { type: 'string' },
```
gnarea, line 7: I just came across https://json-schema.org/understanding-json-schema/reference/non_json_data.html. 

If our dependency supports it, it'd give us a base64-decoded `Buffer`. But it may not support it because it's a relatively new feature in JSON Schema.

It may be worth trying:

```
publicKey: { type: 'string', contentEncoding: "base64" },
```

or,

```
publicKey: { type: 'string', contentEncoding: "base64", contentMediaType: "application/vnd.etsi.tsl.der" },
```

src/testUtils/publicKeyGenerator.ts
```diff
@@ -0,0 +1,16 @@
+import { generateKeyPair } from 'node:crypto';
+import { promisify } from 'node:util';
+
+const generateKeyPairAsync = promisify(generateKeyPair);
+export const generatePublicKey = async () => {
```
gnarea, line 5: FYI, I ended up implementing something equivalent in #76 (`generateKeyPair()`), which I think we should use instead: it also outputs a private key, which we'll need in some cases.

src/memberPublicKey.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+          publicKey: publicKey.toString('base64'),
+          oid: TEST_OID,
+        },
+        serviceOptions,
+      );
+
+      requireSuccessfulResult(memberPublicKey);
+      const dbResult = await memberPublicKeyModel.findById(memberPublicKey.result.id);
+      expect(dbResult?.memberId).toStrictEqual(MEMBER_MONGO_ID);
+      expect(dbResult?.publicKey.toString('base64')).toStrictEqual(publicKey.toString('base64'));
```
gnarea, line 58: We should also check the service OID.

src/memberPublicKey.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+      existingConnection: connection,
+    });
+  });
+
+  describe('createMemberPublicKey', () => {
+    test('Member public key should be created', async () => {
+      const memberPublicKey = await createMemberPublicKey(
+        MEMBER_MONGO_ID,
+        {
+          publicKey: publicKey.toString('base64'),
```
gnarea, line 49: We're doing this in a few tests, so I think we should just have a constant for it.

src/memberPublicKey.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+      });
+
+      const result = await deleteMemberPublicKey(PUBLIC_KEY_ID, serviceOptions);
+
+      requireSuccessfulResult(result);
+      const dbResult = await memberPublicKeyModel.findById(memberPublicKey.id);
+      expect(dbResult).not.toBeNull();
+    });
+
+    test('Record deletion errors should be propagated', async () => {
```
gnarea, line 201: We don't need this test here: We only need it in functions that could failed due to duplicated records (e.g., `createOrg()`).

src/memberPublicKey.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+        publicKey,
+      });
+
+      const result = await getMemberPublicKey(invalidMemberKeyId, MEMBER_MONGO_ID, serviceOptions);
+
+      requireFailureResult(result);
+      expect(result.reason).toBe(MemberPublicKeyProblemType.PUBLIC_KEY_NOT_FOUND);
+    });
+
+    test('Record Find errors should be propagated', async () => {
```
gnarea, line 149: We don't need this test here: We only need it in functions that could failed due to duplicated records (e.g., `createOrg()`).

src/memberPublicKey.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+          oid: TEST_OID,
+        },
+        serviceOptions,
+      );
+
+      requireFailureResult(memberPublicKey);
+      expect(memberPublicKey.reason).toBe(MemberPublicKeyProblemType.MALFORMED_PUBLIC_KEY);
+    });
+
+    test('Record creation errors should be propagated', async () => {
```
gnarea, line 80: We don't need this test here: We only need it in functions that could failed due to duplicated records (e.g., `createOrg()`).

src/services/routes/memberPublicKey.routes.ts
```diff
@@ -0,0 +1,119 @@
...
+  required: ['orgName', 'memberId'],
+} as const;
+
+const MEMBER_PUBLIC_KEY_PARAMS = {
+  type: 'object',
+
+  properties: {
+    orgName: { type: 'string' },
+    memberId: { type: 'string' },
+    memberPrivateKey: { type: 'string' },
```
gnarea, line 39: `memberPublicKeyId`

src/services/routes/memberPublicKey.routes.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+        payload,
+      });
+
+      expect(response).toHaveProperty('statusCode', HTTP_STATUS_CODES.BAD_REQUEST);
+      expect(response.json()).toStrictEqual({
+        type: MemberPublicKeyProblemType.MALFORMED_PUBLIC_KEY,
+      });
+    });
+
+    test('Malformed oid should be refused', async () => {
```
gnarea, line 97: "OID", all caps, please. Applies to all acronyms.

src/services/routes/memberPublicKey.routes.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+        },
+      });
+
+      const response = await serverInstance.inject({
+        ...injectionOptions,
+        payload,
+      });
+
+      expect(response).toHaveProperty('statusCode', HTTP_STATUS_CODES.OK);
+      expect(response.headers['content-type']).toStartWith('application/json');
```
gnarea, line 71: This is superfluous when we're already doing `response.json()`

src/memberPublicKey.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+      );
+
+      requireSuccessfulResult(result);
+      expect(result.result).toMatchObject({
+        oid: TEST_OID,
+        publicKey: publicKey.toString('base64'),
+      });
+    });
+
+    test('Non existent id should return non existing error', async () => {
```
gnarea, line 121: "Existent" seems to be grammatically correct, but most of our tests, including new tests in this PR, use "existing", so I think we should stick to it for consistency.

gnarea (changes requested):

src/models/MemberPublicKey.model.ts
```diff
@@ -0,0 +1,12 @@
+import { prop } from '@typegoose/typegoose';
+
+export class MemberPublicKeyModelSchema {
```
gnarea, line 3: This isn't storing the creation date as required in #7. This is important for auditing purposes.

src/services/schema/memberPublicKey.schema.ts
levansuper: Tried and none of this ensures a valid base64 string
```
{ type: 'string', "contentEncoding": "base64",}

{ 
  type: 'string',
     "media": {
        "binaryEncoding": "base64",
      }
}
```

gnarea: The following isn't correct according to the URL above:

```
{ 
  type: 'string',
     "media": {
        "binaryEncoding": "base64",
      }
}
```

levansuper: Found that on stack overflow. Might be an old version

src/memberPublicKey.spec.ts
gnarea: In `createOrg()` we need two tests indeed because there are two code branches in the `catch` block: the error may or may not be caused by a duplicated record.

But this test doesn't check any branch, because there's no branch to test. There's no `catch`.

levansuper: in that case, there are a lot of places where we can remove this. Shall we remove it?

gnarea: Yes, please. I must've missed those in the reviews.

src/memberPublicKey.spec.ts
```diff
@@ -43,32 +47,37 @@ describe('member public key', () => {
...
+      expect(dbResult!.memberId).toStrictEqual(MEMBER_MONGO_ID);
+      expect(dbResult!.serviceOid).toStrictEqual(TEST_SERVICE_OID);
+      expect(dbResult!.publicKey.toString('base64')).toStrictEqual(publicKeyBase64);
+      expect(dbResult!.creationDate).toBeBetween(commandRunTime, new Date());
       expect(mockLogging.logs).toContainEqual(
         partialPinoLog('info', 'Member public key created', {
           id: memberPublicKey.result.id,
         }),
       );
+      jest.useRealTimers();
```
gnarea, line 72: This doesn't have any effect unless you call `jest.useFakeTimers()` beforehand.

gnarea (changes requested):

src/services/schema/memberPublicKey.schema.ts
gnarea: But did you try the second example in my first comment?

gnarea (approved):

Issue #84, opened 2023-04-08
Implement authorisation in existing endpoints


gnarea:
As per the README.

The healthcheck, which isn't part of the RESTful API documented in the README, should continue to support anonymous requests.

PR #89, opened 2023-04-10
refactor: Nest routes and merge `configureFastify()` into `makeServer()`

gnarea:
This is groundwork for #84, as we need to use [Fastify encapsulation](https://www.fastify.io/docs/latest/Reference/Encapsulation/) to enforce authz.

Note that I also had to merge `configureFastify()` into `makeServer()` as we'll also need to more control over plugins (their execution order and the parameters we pass to them). It never really make sense to separate the two in the first place (I copied them from another project where we have multiple servers, so it makes sense there).


src/services/routes/org.routes.ts
```diff
@@ -162,5 +161,5 @@ export default function registerRoutes(
     },
   });
 
-  done();
+  await fastify.register(memberRoutes, { ...opts, prefix: '/orgs/:orgName/members' });
```
levansuper, line 164: I suggest having a dedicated file for the route prefixes. this way it would be way easier to control the prefixes and check errors in case we have any
For example 
```
const ORG_PREFIX = "/orgs"
const MEMBER_PREFIX = "/orgs/:orgName/members"
```

src/services/routes/member.routes.ts
```diff
@@ -185,5 +184,5 @@ export default function registerRoutes(
     },
   });
 
-  done();
+  await fastify.register(memberPublicKeyRoutes, { ...opts, prefix: '/:memberId/public-keys' });
```
levansuper, line 187: Not completely happy with the way this looks. Maybe we could change this a bit and create some kind of chain function that returns the fastify instance. and then we could construct the paths in one place

would look something like this
```
const orgFastify = registerOrgRoutes(fastifyInstance, opts);
const memberFastify = registerMemberRoutes(orgFastify, opts, prefix: '/:memberId/public-keys');
registerPublicKeyRoutes(memberFastify, opts, prefix: 'public-keys');
registerMemberTokenRoutes(memberFastify, opts, prefix: 'public-keys');
```

levansuper (changes requested):
I don't really like the way we are registering the routes in the different files not completely related to them.
example: 

in org we are registering the member routes

src/services/routes/org.routes.ts
gnarea: > this way it would be way easier to control the prefixes and check errors in case we have any

Can you please elaborate on that?

Unless I made a mistake somewhere, those values are only ever used once, so moving them to a separate file adds complexity and I'm not sure I see the benefit(s).

src/services/routes/member.routes.ts
gnarea: > Not completely happy with the way this looks

I think I understand the alternative you're suggesting, but can you please elaborate on what you don't like? I don't know what problem we're trying to solve.

As for the alternative itself, correct me if I'm wrong, but we'd be creating a bunch of functions that would only ever be called once, only to wrap one-liners like this? It feels overly complicated.

gnarea:
> I don't really like the way we are registering the routes in the different files not completely related to them.

I totally agree with you, but as far as I know Fastify doesn't offer another way to do this.

src/services/server.ts
```diff
@@ -1,31 +1,53 @@
...
  *
  * This function doesn't call .listen() so we can use .inject() for testing purposes.
  */
-export async function makeServer(logger?: Logger): Promise<FastifyInstance> {
-  return configureFastify(
-    [fastifyMongoose, fastifyRoutes, notFoundHandler, ...ROUTES],
-    undefined,
-    logger,
-  );
+export async function makeServer(customLogger?: Logger): Promise<FastifyInstance> {
```
levansuper, line 51: Do we need the logger as a parameter in this function?

levansuper (commented):
I think everything is fine just one quesiton.

src/services/server.ts
gnarea: We do, when we need to check the logs in the unit tests

levansuper (approved):

PR #90, opened 2023-04-10
feat: Implement member key import key create route

levansuper:


src/services/schema/memberPublicKey.schema.ts
```diff
@@ -1,11 +1,13 @@
...
 
+import { SERVICE_OID_REGEX } from './validaitons.js';
+
 export const MEMBER_PUBLIC_KEY_SCHEMA = {
   type: 'object',
 
   properties: {
     publicKey: { type: 'string' },
-    serviceOid: { type: 'string', pattern: '^(?:[0-9]+\\.)*[0-9]+$' },
+    serviceOid: { type: 'string', pattern: SERVICE_OID_REGEX },
```
levansuper, line 10: was just a small change so also changed this

src/memberKeyImportToken.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+      expect(dbResult!.memberId).toStrictEqual(MEMBER_MONGO_ID);
+      expect(dbResult!.serviceOid).toStrictEqual(TEST_SERVICE_OID);
+      expect(mockLogging.logs).toContainEqual(
+        partialPinoLog('info', 'Member key import token created', {
+          id: meberKeyImportToken.result.id,
+        }),
+      );
+    });
+
+    test('Multiple tokens with same member id and service oid should be created', async () => {
```
levansuper, line 48: Not sure about the length of the description

levansuper:
The business logic does not really any failure condition and that makes me think if I missed something. Maybe could pay attention to that

src/services/schema/validaitons.ts
gnarea, line 1: File name has a typo. Should be `validations.ts` or `validation.ts` (better).

src/services/server.ts
```diff
@@ -15,6 +16,7 @@ const ROUTES: FastifyPluginCallback<RouteOptions>[] = [
   orgRoutes,
   memberRoutes,
   memberPublicKey,
+  memberKeyImportToken,
```
gnarea, line 19: Heads-up: Git won't report a merge conflict, but if #89 is merged first, things will break unless we manually move this to `member.routes.ts`.

src/memberKeyImportToken.spec.ts
gnarea: Agreed. How about `Multiple tokens for same member and service should be allowed`?

gnarea (changes requested):
LGTM, just some minor issues spotted.

gnarea:
> The business logic does not really any failure condition and that makes me think if I missed something

I can't think of anything. The only things that could fail here related to the schema or authorisation, which are/will be handled separately.

src/services/schema/validaiton.ts
gnarea, line 1: This file name is still wrong. Notice the typo.

src/services/server.ts
gnarea, line 1: Merge gone wrong? This file shouldn't have changed.

gnarea (changes requested):

gnarea (approved):
Thanks

PR #92, opened 2023-04-11
refactor: Prepare to support multiple HTTP servers

gnarea:
So that we can run Knative Eventing sinks.

# High-level changes

- Rename `src/services` to `src/api`.
- Move `src/api/schema` to `src/schemas` because these things aren't specific to the API server. We may need them in a background queue.
- Move Fastify-related, yet API server-agnostic, utils from `src/api/*` to `src/utils/fastify/*`.
- Reimplement Docker entrypoint so that the image is now run with a single argument: The name of the server to run. Right now only `api` is supported.


levansuper (approved):
Nice. This will make adding new servers way easier!

Issue #79, opened 2023-04-06
Internal server error response includes exception message


gnarea:
I discovered this whilst testing #76.

Example:

```
HTTP/1.1 500 Internal Server Error
content-length: 130
content-type: application/json; charset=utf-8
date: Thu, 06 Apr 2023 22:39:22 GMT
x-envoy-upstream-service-time: 46
server: envoy

{
  "statusCode": 500,
  "error": "Internal Server Error",
  "message": "env-var: \"KMS_ADAPTER\" is a required variable, but it was not set"
}
```

PR #94, opened 2023-04-11
fix: Implement error handler and logger

levansuper:
fixes: #79


src/utilities/fastify/server.ts
```diff
@@ -40,9 +41,21 @@ export async function makeFastify(appPlugin: FastifyPluginAsync, customLogger?:
 
   await server.register(fastifyMongoose);
 
+  const internalServerError = 'Internal server error';
+  server.setErrorHandler(async (error, _request, reply) => {
+    if (error.statusCode !== undefined && error.statusCode < HTTP_STATUS_CODES.INTERNAL_ERROR) {
+      await reply.send(error);
+      return;
+    }
```
levansuper, line 49: Had to implement this check. 
This hook catches all the errors including 4xx

src/api/routes/error.routes.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+    expect(mockLogging.logs).toContainEqual(
+      expect.objectContaining({
+        level: pino.levels.values.info,
+        msg: 'Internal server error',
+
+        err: expect.objectContaining({
+          message: 'ERROR_MESSAGE',
+          stack: expect.toStartWith('Error: '),
+        }),
+      }),
```
gnarea, line 33: Please check out how we use the utility `partialPinoLog()`.

src/testUtils/server.ts
```diff
@@ -1,10 +1,12 @@
 import { makeServer } from '../api/server.js';
 import type { FastifyTypedInstance } from '../utilities/fastify/FastifyTypedInstance.js';
 
-export function setUpTestServer(): () => FastifyTypedInstance {
+import type { MockLogging } from './logging.js';
+
+export function setUpTestServer(mockLogging?: MockLogging): () => FastifyTypedInstance {
```
gnarea, line 6: I think we should just pass the logger for simplicity.

src/utilities/fastify/server.ts
```diff
@@ -40,9 +41,21 @@ export async function makeFastify(appPlugin: FastifyPluginAsync, customLogger?:
...
+
+    logger.info(error, internalServerError);
+    await reply.status(HTTP_STATUS_CODES.INTERNAL_ERROR).send(internalServerError);
```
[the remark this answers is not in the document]
gnarea: Yes, definitely error for the `500` response.

src/utilities/fastify/server.ts
gnarea: Makes sense, although it's a shame Fastify doesn't make it easier.

src/utilities/fastify/server.ts
```diff
@@ -40,14 +41,21 @@ export async function makeFastify(appPlugin: FastifyPluginAsync, customLogger?:
 
   await server.register(fastifyMongoose);
 
-  server.setErrorHandler(function (_error, _request, reply) {
-    reply.status(HTTP_STATUS_CODES.INTERNAL_ERROR).send("Internal server error");
-  })
+  const internalServerError = 'Internal server error';
+  server.setErrorHandler(async (error, _request, reply) => {
+    if (error.statusCode !== undefined && error.statusCode < HTTP_STATUS_CODES.INTERNAL_ERROR) {
+      await reply.send(error);
```
gnarea, line 47: Since we're overriding the default error handler, we should also log an info log. See: https://www.fastify.io/docs/latest/Reference/Server/#seterrorhandler

gnarea (changes requested):
I really think that we should absolutely not be defining an endpoint that intentionally throws errors in production. I don't think it makes sense, but most importantly: remember that we get alerted to errors, so I wouldn't want to be woken up in the middle of the night because someone hit that endpoint.

We definitely need the function inside `error.routes.ts`, but it should be exclusively used in the unit tests. I suggested that we pass it as the first argument to `makeFastify()` in an unit test in the call we had earlier -- why was that ruled out? I think that's the only place where that function belongs -- note how the error handler is defined in `src/utilities/fastify/server.ts` and yet not tested in `src/utilities/fastify/server.spec.ts`.

src/testUtils/errorRoutesApp.ts
```diff
@@ -0,0 +1,48 @@
...
+    },
+  });
+
+  done();
+}
+
+export async function errorAppPlugin(server: FastifyInstance) {
+  await server.register(fastifyRoutes);
+  await server.register(registerErrorRoutes);
+}
```
levansuper, line 48: We could also merge this whole file into the test itself

gnarea: Agreed. This isn't a _utility_ in the sense we use it here, as it isn't (and shouldn't be) shared by multiple test suites.

But more importantly, I don't think we should be defining one plugin for both tests. Unit tests should minimise shared fixtures and that can be easily be accomplished here. For example (untested code, based on the [docs](https://www.fastify.io/docs/latest/Reference/Errors/)):

```ts
import { errorCodes } from 'fastify';

function registerUncaughtErrorRoute(
  fastify: FastifyTypedInstance,
  _opts: RouteOptions,
  done: PluginDone,
): void {
  fastify.route({
    method: ['POST'],
    url: '/uncaught-error',

    handler(): void {
      throw new Error('ERROR_MESSAGE');
    },
  });

  done();
}


function registerClientErrorRoute(
  fastify: FastifyTypedInstance,
  _opts: RouteOptions,
  done: PluginDone,
): void {
  fastify.route({
    method: ['POST'],
    url: '/4xx',

    handler(): void {
      throw new errorCodes.FST_ERR_NOT_FOUND();
      // either that, or the following. Not sure which one should be used here.
      // reply.code(404).send(new errorCodes.FST_ERR_NOT_FOUND())
    },
  });

  done();
}
````

Note how simple the 4XX route can be. I don't think we should be introducing a dependency on something external like `'@fastify/routes'` -- especially when that approach causes unreachable code like the current implementation (the handler is never called).

File: src/utilities/fastify/server.error.spec.ts
Comment (gnarea) [inline src/utilities/fastify/server.error.spec.ts:1]: Having this separate file is inconsistent with our conventions. Tests should go in the respective `describe` inside `server.spec.ts`.

File: src/utilities/fastify/server.error.spec.ts [hunk 7]
```diff
@@ -0,0 +1,56 @@
...
+
+  test('Thrown error should be handled gracefully and logged', async () => {
+    const response = await serverInstance.inject({ method: 'POST', url: '/5xx' });
+    expect(response.headers['content-type']).toStartWith('text/plain');
+    expect(response.body).toBe('Internal server error');
+    expect(mockLogging.logs).toContainEqual(
+      partialPinoLog('error', 'Internal server error', {
+        err: expect.objectContaining({
+          message: 'ERROR_MESSAGE',
+          stack: expect.toStartWith('Error: '),
```
gnarea, line 32:
```suggestion
          type: 'Error',
```

For simplicity.

src/utilities/fastify/server.error.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+      url: '/4xx',
+      payload: {},
+    });
+
+    expect(response).toHaveProperty('statusCode', HTTP_STATUS_CODES.BAD_REQUEST);
+    expect(mockLogging.logs).toContainEqual(
+      partialPinoLog('info', 'Client error', {
+        err: expect.objectContaining({
+          message: "body must have required property 'test'",
+          stack: expect.toStartWith('Error: '),
```
gnarea, line 51:
```suggestion
          type: 'Error',
```

For simplicity.

src/utilities/fastify/server.ts
```diff
@@ -43,13 +43,17 @@ export async function makeFastify(appPlugin: FastifyPluginAsync, customLogger?:
...
+      error.statusCode < HTTP_STATUS_CODES.INTERNAL_SERVER_ERROR
+    ) {
+      logger.info(error, 'Client error');
       await reply.send(error);
       return;
     }
 
-    logger.info(error, internalServerError);
-    await reply.status(HTTP_STATUS_CODES.INTERNAL_ERROR).send(internalServerError);
+    logger.error(error, internalServerError);
```
gnarea, line 55: I forgot that Pino requires the error to be called `err`:

```suggestion
    logger.error({err: error}, internalServerError);
```

src/utilities/fastify/server.ts
```diff
@@ -43,13 +43,17 @@ export async function makeFastify(appPlugin: FastifyPluginAsync, customLogger?:
 
   const internalServerError = 'Internal server error';
   server.setErrorHandler(async (error, _request, reply) => {
-    if (error.statusCode !== undefined && error.statusCode < HTTP_STATUS_CODES.INTERNAL_ERROR) {
+    if (
+      error.statusCode !== undefined &&
+      error.statusCode < HTTP_STATUS_CODES.INTERNAL_SERVER_ERROR
+    ) {
+      logger.info(error, 'Client error');
```
gnarea, line 50: I just remembered that Pino requires errors to be called `err`:

```suggestion
      logger.info({err: error}, 'Client error');
```

gnarea (changes requested):

src/utilities/fastify/handler/setErrorHandler.ts
```diff
@@ -0,0 +1,20 @@
...
+    if (
+      error.statusCode !== undefined &&
+      error.statusCode < HTTP_STATUS_CODES.INTERNAL_SERVER_ERROR
+    ) {
+      fastify.log.info({ err: error }, 'Client error');
+      await reply.send(error);
+      return;
+    }
+
+    fastify.log.error({ err: error }, internalServerError);
```
gnarea, line 17: I don't think the log message should be "Internet server error" here: We don't know if it's a `500`. It could be any other `5XX` error.

I think we should use something like `Request processing failed due to server error`.

It does, however, make sense to return a response body of `Internal server error` to the client. So let's keep that part -- without the variable `internalServerError` as it won't be necessary anymore.

gnarea, line 17: You should always use `request.log` where possible as it contains important context (e.g., request id), which `fastify.log` lacks (because it's agnostic of any particular request).

src/utilities/fastify/handler/setErrorHandler.ts
```diff
@@ -0,0 +1,20 @@
...
+    await reply.status(HTTP_STATUS_CODES.INTERNAL_SERVER_ERROR).send(internalServerError);
```
gnarea, line 18: I think we should use `error.statusCode` if it exists. If it doesn't exist, use `HTTP_STATUS_CODES.INTERNAL_SERVER_ERROR`.

src/utilities/fastify/handler/setErrorHandler.ts
```diff
@@ -0,0 +1,20 @@
...
+import { HTTP_STATUS_CODES } from '../../http.js';
+
+export default function setErrorHandler(fastify: FastifyInstance): void {
+  const internalServerError = 'Internal server error';
+  fastify.setErrorHandler(async (error, _request, reply) => {
+    if (
+      error.statusCode !== undefined &&
+      error.statusCode < HTTP_STATUS_CODES.INTERNAL_SERVER_ERROR
+    ) {
+      fastify.log.info({ err: error }, 'Client error');
```
gnarea, line 12: `request.log` please.

src/utilities/http.ts
```diff
@@ -8,5 +8,7 @@ export const HTTP_STATUS_CODES = {
   NOT_FOUND: 404,
   METHOD_NOT_ALLOWED: 405,
   CONFLICT: 409,
+  UNSUPPORTED_MEDIA_TYPE: 415,
+  FAILED_DEPENDENCY: 424,
```
gnarea: Looks like a bad merge. Are you sure this won't duplicate these lines when merged in `main` eventually?

gnarea (changes requested):

src/utilities/fastify/handler/setErrorHandler.ts
```diff
@@ -0,0 +1,20 @@
+import type { FastifyInstance } from 'fastify';
+
+import { HTTP_STATUS_CODES } from '../../http.js';
+
+export default function setErrorHandler(fastify: FastifyInstance): void {
+  const internalServerError = 'Internal server error';
+  fastify.setErrorHandler(async (error, _request, reply) => {
```
gnarea, line 7: If we use `err` instead of `error` we won't have to rename it in the logs.

gnarea (changes requested):

PR #96, opened 2023-04-12
chore: Implement background queue

gnarea:
# Overview

Implement background queue, with one **example** event producer and consumer for illustration purposes only.

# High-level changes

- Integrate Knative Eventing.
- Introduce **temporary example** in API server (`POST /example-event-publisher`) to illustrate how to use this in #97.
- Make lots of seemingly-irrelevant tweaks to existing Kubernetes/Knative resources (`k8s/*.yml`) to make this work.
- Since this PR introduces a second server, I had to split the `setUpTestServer()` test utility into two functions (one for each server).

# Testing

The event producer and consumer can be tested very easily:

```http
### Example event producer
POST http://veraid-authority.default.10.103.177.106.sslip.io/example-event-publisher
```

`skaffold dev` will then output the following log from the `queue` container:

```
[queue] {"level":30,"time":1681416127681,"pid":1,"hostname":"veraid-authority-queue-00001-deployment-5cc9849665-fq4q7","reqId":"8c5f3880-aa2e-414e-a3ba-8d8bb3e55ca0","event":{"id":"id","time":"2023-04-13T20:02:07.645Z","type":"net.veraid.orgs.creation","source":"https://veraid.net/authority/api","specversion":"1.0","datacontenttype":"application/json; charset=utf-8","subject":"bbc.com","data":{"foo":"bar"},"knativearrivaltime":"2023-04-13T20:02:07.656158605Z"},"msg":"Event processed"}
```

Which means that the event produced in the API server reached the right sink in the background queue.

### DIFF
--- a/src/backgroundQueue/QueueProblemType.ts
+++ b/src/backgroundQueue/QueueProblemType.ts
@@ -1,4 +1,4 @@
-export enum ProblemType {
+export enum QueueProblemType {
   INVALID_EVENT = 'https://veraid.net/problems/internal-events/invalid-event',
   UNSUPPORTED_EVENT = 'https://veraid.net/problems/internal-events/unsupported-event',
 }
--- a/src/backgroundQueue/server.spec.ts
+++ b/src/backgroundQueue/server.spec.ts
@@ -7,7 +7,7 @@ import { HTTP_STATUS_CODES } from '../utilities/http.js';
 import { CE_ID, CE_SOURCE } from '../testUtils/eventing/stubs.js';
 import { postEvent } from '../testUtils/eventing/cloudEvents.js';
 
-import { ProblemType } from './ProblemType.js';
+import { QueueProblemType } from './QueueProblemType.js';
 
 configureMockEnvVars(REQUIRED_QUEUE_ENV_VARS);
 
@@ -38,7 +38,7 @@ describe('makeQueueServer', () => {
       });
 
       expect(response.statusCode).toBe(HTTP_STATUS_CODES.BAD_REQUEST);
-      expect(response.json()).toHaveProperty('type', ProblemType.INVALID_EVENT);
+      expect(response.json()).toHaveProperty('type', QueueProblemType.INVALID_EVENT);
     });
 
     test('Unsupported CloudEvent type should be refused', async () => {
@@ -52,7 +52,7 @@ describe('makeQueueServer', () => {
       const response = await postEvent(event, server);
 
       expect(response.statusCode).toBe(HTTP_STATUS_CODES.BAD_REQUEST);
-      expect(response.json()).toHaveProperty('type', ProblemType.UNSUPPORTED_EVENT);
+      expect(response.json()).toHaveProperty('type', QueueProblemType.UNSUPPORTED_EVENT);
     });
   });
 });
--- a/src/backgroundQueue/server.ts
+++ b/src/backgroundQueue/server.ts
@@ -9,7 +9,7 @@ import { EXAMPLE_EVENT_TYPE } from '../internalEvents/example.js';
 
 import processExample from './sinks/example.sink.js';
 import type { Sink } from './Sink.js';
-import { ProblemType } from './ProblemType.js';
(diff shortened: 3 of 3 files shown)


src/utilities/mongo.ts
```diff
@@ -1,8 +1,8 @@
 import envVar from 'env-var';
 import { type Connection, createConnection } from 'mongoose';
 
-export async function createMongooseConnectionFromEnv(): Promise<Connection> {
+export function createMongooseConnectionFromEnv(): Connection {
   const mongoUri = envVar.get('MONGODB_URI').required().asString();
 
-  return createConnection(mongoUri).asPromise();
```
gnarea, line 7: This is the ultimate source of the pain I experienced with this task.

Basically, the original version was blocking the start of the server until the DB connection was established, but Knative expects the server to start listening ASAP, even if it's not ready (even if the health check will time out or return 50X responses).

Note that the both the old and new version of this function returns a Mongoose `Connection` instance. Any DB query will still hang until the connection is established or the timeout is hit. So, exactly the same behaviour as before.

src/utilities/knativeEventing.ts
```diff
@@ -0,0 +1,8 @@
+import { emitterFor, type EmitterFunction, httpTransport } from 'cloudevents';
+import envVar from 'env-var';
+
+export function makeEmitterFromEnv(): EmitterFunction {
```
gnarea, line 4: This function isn't used in this PR, but will be used when we implement `MemberPublicKeyImport` in `POST /awala`. (I tested it by integrating it in the healthcheck, but I'm obviously not committing that change)

src/utilities/eventing/Emitter.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+import { CE_SOURCE, K_SINK } from '../../testUtils/eventing.js';
+import { mockSpy } from '../../testUtils/jest.js';
+
+const mockEmitter = mockSpy(jest.fn());
+const mockTransport = Symbol('mockTransport');
+jest.unstable_mockModule('cloudevents', () => ({
+  emitterFor: jest.fn<any>().mockReturnValue(mockEmitter),
+  httpTransport: jest.fn<any>().mockReturnValue(mockTransport),
+}));
+// eslint-disable-next-line @typescript-eslint/naming-convention
```
gnarea, line 15: Same thing as in `Kms.spec.ts`: This is a workaround until Jest fixes its EMS support and `unstable_mockModule()` is no longer needed.

src/testUtils/logging.ts
```diff
@@ -14,14 +14,13 @@ export function makeMockLogging(): MockLogging {
...
-  return { logger, logs };
-}
 
-// eslint-disable-next-line import/no-unused-modules
-export function partialPinoLogger(bindings: { readonly [key: string]: any }): any {
-  return expect.objectContaining({
-    [PinoSymbols.formattersSym]: { bindings },
+  beforeEach(() => {
+    // Clear the logs
+    logs.splice(0, logs.length);
```
gnarea, line 20: I accidentally wrote a test that passed because of a log that was recorded in a previous test, so that's why I'm forcing the set of logs to be cleared between tests.

src/backgroundQueue/ProblemType.ts
```diff
@@ -0,0 +1,4 @@
+export enum ProblemType {
```
levansuper, line 1: Maybe call this QueueProblemType?

gnarea: I considered that, but this is inside `src/backgroundQueue` and there can't be further problem types.

gnarea: Note that we have to return the same HTTP response, even if the sink refuses the input.

gnarea: Actually, on reflection, it'd indeed be better to call it that, for when one looks it up on the IDE.

gnarea: Change made

levansuper (approved):

PR #99, opened 2023-04-14
feat: Implement Awala endpoint with member bundle request

levansuper:


src/api/routes/awala.routes.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+        url: '/awala',
+        headers: validHeaders,
+
+        payload: {
+          ...validPayload,
+          signature: 'INVALID_BASE_64',
+        },
+      });
+
+      expect(response).toHaveProperty('statusCode', HTTP_STATUS_CODES.BAD_REQUEST);
```
gnarea, line 114: None of these tests are checking the reason (e.g., the problem type), so for all we know they could be passing because the content type doesn't match the supported one(s).

src/schemas/validation.ts
```diff
@@ -1 +1,2 @@
 export const SERVICE_OID_REGEX = '^(?:[0-9]+\\.)*[0-9]+$';
+export const BASE_64 = '^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$';
```
gnarea, line 2: I think we should use the suffix "regex" for consistency.

src/api/routes/awala.routes.ts
```diff
@@ -0,0 +1,85 @@
...
+  schema: Schema,
+): ValidateTypeResult<Schema> {
+  if (!ajv.validate(schema, value)) {
+    return undefined;
+  }
+
+  return value as ValidateTypeResult<Schema>;
+}
+
+async function processMemberBundleRequest(
```
gnarea, line 29: Let's add a log for each of the two branches in this function.

gnarea (changes requested):

src/api/routes/awala.routes.ts
```diff
@@ -0,0 +1,96 @@
...
+import type { PluginDone } from '../../utilities/fastify/PluginDone.js';
+import type { FastifyTypedInstance } from '../../utilities/fastify/FastifyTypedInstance.js';
+import { MEMBER_BUNDLE_REQUEST_SCHEMA } from '../../schemas/awala.schema.js';
+import { AwalaProblemType } from '../../AwalaProblemType.js';
+import { createMemberBundleRequest } from '../../awala.js';
+import type { ServiceOptions } from '../../serviceTypes.js';
+
+const ajv = addFormats(new Ajv());
+
+type ValidationResult<Schema extends JSONSchema> = FromSchema<Schema> | string;
```
levansuper, line 16: Hate to do this but there is something wrong with Typescript if I do `Result<FromSchema<Schema> , string>` it throws an error
Screenshot 2023-04-17 at 15 43 07

src/api/routes/awala.routes.ts
```diff
@@ -0,0 +1,96 @@
...
+      const contentType = awalaRequestMessageTypeList.find(
+        (messageType) => messageType === request.headers['content-type'],
+      );
+
+      const serviceOptions = {
+        logger: this.log,
+        dbConnection: this.mongoose,
+      };
+
+      const result = await awalaEventToProcessor[contentType!](request.body, serviceOptions);
```
levansuper, line 85: All the other content types are rejected by "addContentTypeParser" so this will be a valid one 100%

src/api/routes/awala.routes.ts
gnarea: The change to `Result<FromSchema<Schema> , string>` makes sense to me and I can't see anything wrong with your code.

An option we considered last week was to throw a custom error here (custom as in a subclass of `Error`)? That'd get around this issue -- it'd be a pragmatic solution, certainly not the nicest one.

I'm just thinking this is a tiny function that won't be exposed outside this file and we've already spent a lot time trying to find a nice solution for it (first with the assertions), so at this point we should just go for something that works.

src/api/routes/awala.routes.ts
```diff
@@ -0,0 +1,96 @@
...
+  return value as ValidationResult<Schema>;
+}
+
+async function processMemberBundleRequest(
+  data: unknown,
+  options: ServiceOptions,
+): Promise<AwalaProblemType | undefined> {
+  const validationResult = validateMessage(data, MEMBER_BUNDLE_REQUEST_SCHEMA);
+  if (typeof validationResult === 'string') {
+    options.logger.info(data, validationResult);
```
gnarea, line 35: `data` contains a couple of binary, base64-encoded fields (the key will be under 1kib, but the Awala PDA could be anywhere between 1kib and 4kib), so it's too much data to send to the logging service. 

Also, each log entry should have a static message so we can easily locate the log whilst debugging, so the `validationResult` (which is dynamic) should be part of the log data instead.

```suggestion
    options.logger.info(
      {publicKeyId: (data as any).publicKeyId, reason: validationResult},
       'Refused invalid member bundle request',
     );
```

The `publicKeyId` will be handy when troubleshooting, to figure out whom needs to be contacted.

src/api/routes/awala.routes.ts
```diff
@@ -0,0 +1,96 @@
...
+    return ajv.errorsText(ajv.errors);
+  }
+
+  return value as ValidationResult<Schema>;
+}
+
+async function processMemberBundleRequest(
+  data: unknown,
+  options: ServiceOptions,
+): Promise<AwalaProblemType | undefined> {
```
gnarea, line 32: It doesn't look idiomatic to do use `Promise<undefined>` and its corresponding `return undefined`. Doesn't the following work?

```suggestion
): Promise<AwalaProblemType | void> {
```

src/api/routes/awala.routes.ts
gnarea: Agreed.

gnarea, line 85: This statement could be a litter simpler, which would also make it easier to debug:

```suggestion
      const processor = awalaEventToProcessor[contentType!];
      const result = await processor(request.body, serviceOptions);
```

src/api/routes/awala.routes.ts
```diff
@@ -0,0 +1,96 @@
...
+      );
+
+      const serviceOptions = {
+        logger: this.log,
+        dbConnection: this.mongoose,
+      };
+
+      const result = await awalaEventToProcessor[contentType!](request.body, serviceOptions);
+
+      if (!result) {
```
gnarea, line 87: We're still not actually using the contents of `result` as mentioned in my initial review. `result` is only ever used as a boolean, so the problem type is being ignored.

gnarea (changes requested):

PR #103, opened 2023-04-17
fix(API): Implement authorisation

### LINKED ISSUE #84: Implement authorisation in existing endpoints

gnarea:
Fixes #84.

# High-level changes

- Implement _super admin_ support by using an optional env var (`AUTHORITY_SUPERADMIN`) that specifies the JWT subject id of the super admin (i.e., their email address).
- **Existing API tests are now run as a super admin**.
- Update test suite for each endpoint to run authorisation checks with every type of user (using test util `testOrgRouteAuth()`).
- Anonymous users now get a `401` on any endpoint under `/orgs`.
- Regular members aren't allowed to update their own membership for security reasons: they shouldn't be allowed to change anything -- especially their user name. 
- Regular members aren't allowed to delete their own membership: I can't think of any legitimate use case for it, but I can think of things goes awry if the membership dataset for an organisation goes out of sync with the org's data sources.

# Notes

- Auth takes precedence over existence checks for security reasons. So, for example, no authenticated user could just request `GET /orgs/bbc.com` to check if `bbc.com` is registered:
  - A `403` will be returned regardless of whether `bbc.com` exists or not, if the current user isn't an admin of `bbc.com`. For example, if the user is a regular member of `bbc.com` or not even a member at all.
  - A `200` will be returned if the org exists and the user is an admin of `bbc.com`.
  - Super admins will get a `200` if the org exists or `404` if it doesn't.

# Testing

First off, you need an access token to make authenticated requests to anything under `/orgs`. To obtain one for the super admin (`admin@veraid.example`), run:

```http
### Authenticate with authorisation server (client credentials)
POST http://mock-authz-server.default.10.103.177.106.sslip.io/default/token
Content-Type: application/x-www-form-urlencoded

grant_type=client_credentials&client_id=admin@veraid.example&client_secret=s3cr3t
```

Replace the email address above to impersonate any user -- even someone who's not a member of any org.

You can now make authenticated requests. For example:

```http
### Create org
POST http://veraid-authority.default.10.103.177.106.sslip.io/orgs
Authorization: Bearer <INSERT-ACCESS-TOKEN-HERE>
Content-Type: application/json

{
  "name": "example.com",
  "memberAccessType": "OPEN"
}

### Get org
GET http://veraid-authority.default.10.103.177.106.sslip.io/orgs/example.com
Authorization: Bearer <INSERT-ACCESS-TOKEN-HERE>
```


src/testUtils/apiServer.ts
```diff
@@ -1,20 +1,259 @@
...
+  });
+}
+
+function unsetAuthUser(fastify: FastifyInstance) {
+  getMockAuthenticateFromServer(fastify).mockImplementation(async (_request, reply) => {
+    await reply.code(HTTP_STATUS_CODES.UNAUTHORIZED).send();
+  });
+}
+
+type RouteLevel = 'ORG_BULK' | 'ORG_MEMBERSHIP_RESTRICTED' | 'ORG_MEMBERSHIP' | 'ORG';
```
gnarea, line 88: Tried to use an enum first, but because this file has to be imported with `const {RouteLevel} = await import(...)`, that'd break the naming ESLint rule because `RouteLevel` would be a variable and as such it can't be PascalCased.

src/api/orgAuthPlugin.ts
```diff
@@ -0,0 +1,116 @@
...
+  readonly reason: string;
+}
+
+async function decideAuthorisation(
+  userEmail: string,
+  request: FastifyRequest,
+  dbConnection: Connection,
+  superAdmin?: string,
+): Promise<Result<AuthorisationGrant, string>> {
+  if (superAdmin === userEmail) {
```
levansuper, line 36: Maybe we can set multiple super admin accounts? We could have an array?

src/api/orgAuthPlugin.ts
```diff
@@ -0,0 +1,116 @@
...
+    return { didSucceed: false, reason: 'Non-super admin tries to access bulk org endpoint' };
+  }
+
+  const memberModel = getModelForClass(MemberModelSchema, {
+    existingConnection: dbConnection,
+  });
+  const member = await memberModel.findOne({ orgName, memberId }).select(['role', 'email']);
+  if (member === null) {
+    return { didSucceed: false, reason: 'User is not a member of the org' };
+  }
```
levansuper, line 52: Can't an org admin "GET" an org?

src/api/orgAuthPlugin.ts
```diff
@@ -0,0 +1,116 @@
...
+  if (member.role === Role.ORG_ADMIN) {
+    return { didSucceed: true, result: { reason: 'User is org admin', isAdmin: true } };
+  }
+
+  if (memberId === undefined) {
+    return { didSucceed: false, reason: 'User is not accessing a membership' };
+  }
```
levansuper, line 59: I think we need to check this before we even try to select the members form the database

src/api/orgAuthPlugin.ts
gnarea: They can. Access would've been granted above.

src/api/orgAuthPlugin.ts
gnarea: I considered that but ruled it out: Although it certainly seems like the more intuitive option, it would complicate the code because we'd need to introduce an extra `if`/`else` to make sure org admins are still given access to anything under `/orgs/:orgName` even if there's no `:memberId`.

src/api/orgAuthPlugin.ts
gnarea: Sorry, I mean *below*.

src/api/orgAuthPlugin.ts
gnarea: Also, because we're delegating authentication to a third party, operators could easily configure their auth server to issue access tokens from different sets of user credentials.

For example, they could have one set of credentials to access the API programmatically via the JS library and another set of credentials to access the API manually via a CLI. Both sets of credentials would map to the same email address.

src/api/orgAuthPlugin.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  fastify.decorateRequest('isUserAdmin', false);
+
+  fastify.addHook('onRequest', async (request, reply) => {
+    const superAdmin = envVar.get('AUTHORITY_SUPERADMIN').asString();
+    const userEmail = (request as AuthenticatedFastifyRequest).user.sub;
+    const decision = await decideAuthorisation(userEmail, request, fastify.mongoose, superAdmin);
+    const reason = decision.didSucceed ? decision.result.reason : decision.reason;
+    if (decision.didSucceed) {
+      (request as AuthorisedFastifyRequest).isUserAdmin = decision.result.isAdmin;
```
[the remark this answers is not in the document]
levansuper: correct. isUserAdmin just sounds odd to me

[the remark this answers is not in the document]
gnarea: Gotcha 👍🏾 

The correct English prose would be "Is user an admin?", so I think that name isn't too far off, considering we omit articles (e.g., "a", "an") from names.

levansuper (approved):

Issue #107, opened 2023-04-19
Implement member bundle request scheduler


gnarea:
I already implemented the scaffolding around this in https://github.com/relaycorp/veraid-authority/blob/1ca6f08f5cf18480f969eb7253b5d988f3fd6c8b/src/backgroundQueue/sinks/memberBundleRequestTrigger.sink.ts

We just need to update that file with the actual business logic:

1. Find all the member bundle requests in the database with an start date in the next 3 days.
2. Fire the `net.veraid.authority.member-bundle-request` event for each record and then delete the DB record. Make sure to use the `MemberBundleRequest.id` as the CloudEvent `id`, to allow the broker to dedupe events.

Keep in mind that this can be tested with `skaffold dev`: That sink is triggered every minute.

PR #112, opened 2023-04-21
feat: Implement process member bundle request queue

levansuper:
closes: #107


src/backgroundQueue/sinks/example.sink.ts
```diff
@@ -1,12 +1,12 @@
...
 // eslint-disable-next-line @typescript-eslint/require-await
 export default async function processExample(
   event: CloudEvent<unknown>,
-  logger: BaseLogger,
+  options: ServiceOptions,
 ): Promise<void> {
   // NB: A production-worthy implementation would do schema validation on `event.data` here.
 
-  logger.info({ event: event.toJSON() }, 'Event processed');
+  options.logger.info({ event: event.toJSON() }, 'Event processed');
```
gnarea, line 11: Make sure the final event log doesn't include the whole event -- just the few things we'd need to identify the event whilst troubleshooting.

src/backgroundQueue/sinks/memberBundleRequestTrigger.sink.spec.ts
gnarea, line 1: We should also test the case where there are no requests to process (no event should be emitted, no DB record deleted).

src/backgroundQueue/sinks/memberBundleRequestTrigger.sink.spec.ts
```diff
@@ -33,14 +61,100 @@ describe('triggerBundleRequest', () => {
...
+    const triggerEvent = new CloudEvent<MemberBundleRequestTriggerPayload>({
+      id: CE_ID,
+      source: CE_SOURCE,
+      type: BUNDLE_REQUEST_TRIGGER_TYPE,
+    });
+
+    await postEvent(triggerEvent, server);
+
+    const publishedEvents = getEvents();
+    expect(publishedEvents).toHaveLength(1);
```
gnarea, line 110: This seems to be outside the scope of this test, and is already tested above.

src/backgroundQueue/sinks/memberBundleRequestTrigger.sink.spec.ts
```diff
@@ -33,14 +61,100 @@ describe('triggerBundleRequest', () => {
...
+  });
+
+  test('Bundle start date more then 3 days into the future should not be sent', async () => {
+    const data = {
+      publicKeyId: MEMBER_PUBLIC_KEY_MONGO_ID,
+      memberBundleStartDate: addSeconds(addDays(new Date(), BUNDLE_REQUEST_DATE_RANGE), 1),
```
gnarea, line 143: A 1-second difference is too small, so this is very likely to be a flaky test. I've been bitten by this too many times on CI, where performance is highly unpredictable 😬 

Let's increase to something like 5 or 10 seconds, which is still a negligible number considering where's dealing with days.

gnarea (changes requested):

src/backgroundQueue/sinks/memberBundleRequestTrigger.sink.ts
```diff
@@ -1,29 +1,56 @@
...
+import type { ServiceOptions } from '../../serviceTypes.js';
 
+export const BUNDLE_REQUEST_DATE_RANGE = 3;
 export default async function triggerBundleRequest(
   event: CloudEvent<unknown>,
-  logger: BaseLogger,
+  options: ServiceOptions,
 ): Promise<void> {
-  logger.info({ source: event.source }, 'Triggering bundle request');
+  options.logger.info({ eventId: event.id, eventType: event.type }, 'Triggering bundle request');
```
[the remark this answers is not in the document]
levansuper: I think we need to log eventId and eventType to identify the event.

PR #114, opened 2023-04-21
fix: Drop support for org-level Awala endpoints

gnarea:
We don't need that ever since we started capturing Awala PDAs.


src/org.spec.ts
```diff
@@ -184,27 +147,6 @@ describe('org', () => {
   });
 
   describe('updateOrg', () => {
-    test('Valid data should be updated', async () => {
```
gnarea, line 187: I deleted this test because there's no data to update. We still check the name is the same in the test below.

levansuper (approved):

src/org.spec.ts
levansuper: in this case we don't we need to remove the endpoint at all?

PR #119, opened 2023-04-24
feat: Implement member bundle issuance

levansuper:


src/awala.ts
```diff
@@ -1,9 +1,24 @@
...
+  retrieveVeraDnssecChain,
+  selfIssueOrganisationCertificate, serialiseMemberIdBundle,
+} from '@relaycorp/veraid';
+import { MemberModelSchema } from './models/Member.model.js';
+import { Kms } from './utilities/kms/Kms.js';
+import env from 'env-var';
+import { OrgModelSchema } from './models/Org.model.js';
+import { derDeserializePublicKey } from './utilities/derDeserialisePublicKey.js';
+
+const awalaEndpoint = env.get('AWALA_MW').asString();
```
gnarea, line 20: I think the word URL is important.

src/utilities/derDeserialisePublicKey.ts
gnarea, line 1: Note the the patch I sent had this implemented in `src/utilities/webcrypto.ts`. I think we should keep it there instead of creating a new file.

src/awala.ts
```diff
@@ -33,3 +48,117 @@ export async function createMemberBundleRequest(
     didSucceed: true,
   };
 }
+// string will be substituted with the actual error
+export async function generateMemberBundle(
```
gnarea, line 52: I don't think this function belongs in this file: It does, indeed, use the Awala middleware, but just like a [_backing service_](https://12factor.net/backing-services) -- much like MongoDB and the Knative broker (emitter).

The functionality it implements is agnostic of the `POST /awala` endpoint. It's more closely related to the _org member_, so I think we should create a new file like `src/memberBundle.ts` -- analogous to all the other business logic files in `src/*.ts`.

src/awala.ts
```diff
@@ -33,3 +48,117 @@ export async function createMemberBundleRequest(
...
+
+  const memberCertificate = await issueMemberCertificate(
+    member.name,
+    memberCryptoPublicKey,
+    orgCertificate,
+    orgPrivateKey,
+    expiryDate,
+  );
+
+  const dnssecChain = await retrieveVeraDnssecChain(member.orgName, undefined);
```
gnarea, line 125: Two things:

- This is quite likely to throw an error, as it connects to a DNS resolver on the Internet, so it'd be good to catch any errors and do a `log.error(...)`. Otherwise, it'd be an uncaught error that would crash the process.
- I _think_ the `undefined` is unnecessary.

gnarea (changes requested):
Overall approach looks good to me.

src/utilities/kms/Kms.ts
```diff
@@ -4,6 +4,16 @@ import { bufferToArrayBuffer } from '../buffer.js';
...
+import { Crypto } from '@peculiar/webcrypto';
+import { CryptoEngine, setEngine } from 'pkijs';
+
+const crypto = new Crypto();
+const cryptoEngine = new CryptoEngine({
+  crypto,
+  name: 'nodeEngine',
+  subtle: crypto.subtle,
+});
+setEngine('nodeEngine', cryptoEngine);
```
levansuper, line 16: For tests, this is not needed anymore. Do we need this in prod?

src/memberBundle.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+      expect(issueMemberCertificateParameters[2]).toBe(selfIssueOrganisationCertificateResult);
+      expect(issueMemberCertificateParameters[3]).toBe(calledKeyPair.privateKey);
+      expect(issueMemberCertificateParameters[4].getTime()).toBe(selfIssueOrganisationCertificateParameters[2].getTime());
+
+      expect(retrieveVeraDnssecChain).toHaveBeenCalledOnceWith(ORG_NAME);
+      expect(serialiseMemberIdBundle).toHaveBeenCalledOnceWith(issueMemberCertificateResult, selfIssueOrganisationCertificateResult, retrieveVeraDnssecChainResult);
+      expect(result.result).toBe(serialiseMemberIdBundleResult);
+
+
+      });
```
levansuper, line 137: The test looks ugly but this way every method is tested. Maybe we can split the test into multiple to make it more readable

src/MemberBundleProblemType.ts
```diff
@@ -0,0 +1,8 @@
+export enum MemberBundleProblemType {
+  MEMBER_NOT_FOUND = 'https://veraid.net/problems/member-not-found',
+  MEMBER_NO_NAME = 'https://veraid.net/problems/member-no-name',
+  ORG_NOT_FOUND = 'https://veraid.net/problems/org-not-found',
+  PUBLIC_KEY_NOT_FOUND = 'https://veraid.net/problems/member-public-key-not-found',
+  BUNDLE_GENERATION_ERROR = 'https://veraid.net/problems/member-bundle-generation-error',
+
+}
```
levansuper, line 8: We could import the errors from the other problem types like:
```
MEMBER_NOT_FOUND = MemberProblemType.MEMBER_NOT_FOUND
```
or we could chenge the method return type:
```
Promise<Result<ArrayBuffer, MemberBundleProblemType>>
```
to
```
Promise<Result<ArrayBuffer, MemberBundleProblemType | MemberProblemType | OrgProblemType>>

```

src/testUtils/veraid.mock.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  beforeEach(() => {
+    mockedModule.issueMemberCertificate.mockReset();
+    mockedModule.retrieveVeraDnssecChain.mockReset();
+    mockedModule.selfIssueOrganisationCertificate.mockReset();
+    mockedModule.serialiseMemberIdBundle.mockReset();
+  });
+
+  return mockedModule;
+}
```
levansuper, line 48: We can move this file to memberBundle.spec.ts but it will create additional noise

gnarea (commented):
Given the state of the code and the fact I don't know what you're planning to improve or not, I only focused on:

- The logic and the security of the functions `createMemberBundleRequest()` and `postToAwala()`.
- The comments you raised.

src/awala.ts
gnarea: I meant we should _add_ the word "URL", not necessarily drop "middleware". Both are key words.

`AWALA_MIDDLEWARE_URL`.

src/memberBundle.ts
```diff
@@ -0,0 +1,109 @@
...
+    }
+  }
+  const member = await memberModel.findById(memberPublicKey.memberId);
+  if(!member){
+    return {
+      didSucceed: false,
+      reason: MemberBundleProblemType.MEMBER_NOT_FOUND
+    }
+  }
+  if(!member.name){
```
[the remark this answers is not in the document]
levansuper: name is required in issueMemberCertificate so I guess we will need something default

src/awala.ts
```diff
@@ -33,3 +33,35 @@ export async function createMemberBundleRequest(
     didSucceed: true,
   };
 }
+
+export async function postToAwala(
+  data: string,
```
levansuper, line 38: What type should data be? Should we allow anything and then stringify or just a string?

src/memberBundle.ts
```diff
@@ -0,0 +1,120 @@
...
+import { OrgModelSchema } from './models/Org.model.js';
+import { derDeserialisePublicKey } from './utilities/webcrypto.js';
+import { Result } from './utilities/result.js';
+
+export async function generateMemberBundle(
+  publicKeyId: string,
+  options: ServiceOptions,
+): Promise<Result<ArrayBuffer, {
+  shouldRetry: boolean
+}>> {
```
levansuper, line 23: I don't really like the response. 
Having Result<ArrayBuffer, boolean> has readability problem when you call the function. So if it fails we get 
```
{
      didSucceed: false,
      reason:true
}
```
And it's really hard to decide what does `reason:true/false` mean

This way it's a bit more visible but still maybe we could create a new type

gnarea: How about rename `FailureResult.reason` to `context`? "Context" is more generic and works in all cases.

levansuper: That will require to rework almost the whole codebase since it's used a lot

src/memberBundle.spec.ts
```diff
@@ -0,0 +1,247 @@
...
+            expect.anything(),
+            expect.anything(),
+          );
+        });
+
+        test('Should be issued with org private key', async () => {
+          await generateMemberBundle(memberPublicKey._id.toString(), serviceOptions);
+
+          // eslint-disable-next-line unicorn/no-unreadable-array-destructuring
+          const [[, , , orgPrivateKey]] = issueMemberCertificate.mock.calls;
```
levansuper, line 180: Since this is a test and the rule is not breaking one I think we can disable it

src/utilities/kms/Kms.ts
levansuper: during `skaffold dev` error is there without this
```
"err":{"type":"Error","message":"Please call 'setEngine' before call to 'getEngine'","stack":"Error: Please call 'setEngine' before call to 'getEngine'\n    at getEngine (/opt/veraid-authority/node_modules/pkijs/build/index.js:6845:19)\n    at getCrypto (/opt/veraid-authority/node_modules/pkijs/build/index.js:6852:21)\n    at PublicKeyInfo.importKey (/opt/veraid-authority/node_modules/pkijs/build/index.js:3919:41)\n    at Function.issue (/opt/veraid-authority/node_modules/@relaycorp/veraid/src/lib/utils/x509/Certificate.ts:162:42)\n    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)\n    at selfIssueOrganisationCertificate (/opt/veraid-authority/node_modules/@relaycorp/veraid/src/lib/pki/organisation.ts:11:23)\n    at generateBundle (/opt/veraid-authority/src/memberBundle.ts:51:26)\n    at generateMemberBundle (/opt/veraid-authority/src/memberBundle.ts:150:24)\n    at aaaa (/opt/veraid-authority/src/memberBundle2.ts:70:9)\n    at triggerBundleRequest (/opt/veraid-authority/src/backgroundQueue/sinks/memberBundleRequestTrigger.sink.ts:72:3)"},"msg":"Please call 'setEngine' before call to 'getEngine'"}
```

gnarea (changes requested):
I reviewed all the code, except for the test files as I'm aware that's work in progress.

Code looks good to me overall. 

I also looked at the new structure of `src/memberBundle.spec.ts`, which looks good to me now.

src/memberBundle.spec.ts
```diff
@@ -0,0 +1,247 @@
...
+        await generateMemberBundle(memberPublicKey._id.toString(), serviceOptions);
+
+        expect(serialiseMemberIdBundle).toHaveBeenCalledOnceWith(
+          issueMemberCertificateResult,
+          expect.anything(),
+          expect.anything(),
+        );
+      });
+
+      describe('Member bundle serialisation', () => {
```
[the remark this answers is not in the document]
levansuper: Moved the top one but the bottom one tests the `generateMemberBundle` output not the "Member bundle serialisation"

src/memberBundle.spec.ts
```diff
@@ -0,0 +1,247 @@
...
+          expect(memberCertificateBuffer).toStrictEqual(memberPublicKeyBuffer);
+        });
+
+        test('Should be issued with org self signed certificate', async () => {
+          await generateMemberBundle(memberPublicKey._id.toString(), serviceOptions);
+
+          expect(issueMemberCertificate).toHaveBeenCalledOnceWith(
+            expect.anything(),
+            expect.anything(),
+            selfIssueCertificateResult,
```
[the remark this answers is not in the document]
levansuper: Did the test and you are right. Will substitute the checks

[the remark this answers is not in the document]
levansuper: But this works so will leave it when we are testing the results
```
        expect(result.result).toStrictEqual(new ArrayBuffer(4));
```

src/awala.spec.ts
```diff
@@ -97,4 +101,45 @@ describe('awala', () => {
...
+      })
+      expect(mockFetch).toHaveBeenNthCalledWith(2, TEST_AWALA_ENDPOINT, {
+        method: 'POST',
+        headers: {
+          [contentTypeHeaderName]: 'application/vnd.veraid.member-bundle',
+          [awalaRecipientHeaderName]: TEST_RECIPIENT_ID
+        },
+        body: awalaPostData,
+      })
+    })
```
levansuper, line 132: Since the test is pretty small I made just one test for it. Should we split this into 2 parts? 
1 - should make authorization request
2 - should post data

src/memberBundle.spec.ts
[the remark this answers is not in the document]
gnarea: I'm cool with using `ArrayBuffer`s where they do work, but please do not compare uninitialised `ArrayBuffer`s as that's extremely unreliable (not just with Jest): `ArrayBuffer`s are never supposed to be compared or manipulated like that. This is a long/complicated topic that we don't want to discuss here, but if you're curious, check out: https://javascript.info/arraybuffer-binary-arrays

Instead, make sure to actually populate the `ArrayBuffer`s. This is what I've done in other repos:

1. Create/update `src/testUtils/buffer.ts` and add the following:
   ```ts
   function bufferToArrayBuffer(buffer: Uint8Array): ArrayBuffer {
     return buffer.buffer.slice(buffer.byteOffset, buffer.byteOffset + buffer.byteLength);
   }

   export function stringToArrayBuffer(string: string): ArrayBuffer {
     return bufferToArrayBuffer(Buffer.from(string));
   }
   ```
2. Initialise every `ArrayBuffer` in the test suite; e.g.:
   ```ts
selfIssueCertificateResult = stringToArrayBuffer('org certificate');
   ```

src/awala.spec.ts
gnarea: Yes, please. I think it should be split. I should be able to understand the requirements from the test suite.

package.json
```diff
@@ -34,7 +35,8 @@
     "is-valid-domain": "^0.1.6",
     "json-schema-to-ts": "^2.8.0",
     "mongoose": "^7.0.4",
-    "pino": "^8.11.0"
+    "pino": "^8.11.0",
+    "pkijs": "^3.0.14"
```
gnarea, line 39: `pkijs` and all other code related to the workaround should be removed once I've fixed the root issue.

src/memberBundle.ts
```diff
@@ -0,0 +1,175 @@
...
+import type { ServiceOptions } from './serviceTypes.js';
+import { MemberPublicKeyModelSchema } from './models/MemberPublicKey.model.js';
+import { MemberModelSchema } from './models/Member.model.js';
+import { Kms } from './utilities/kms/Kms.js';
+import { OrgModelSchema } from './models/Org.model.js';
+import { derDeserialisePublicKey } from './utilities/webcrypto.js';
+import type { Result } from './utilities/result.js';
+
+const CERTIFICATE_EXPIRY_DAYS = 90;
+interface CreateBundleInput {
```
gnarea, line 20:
```suggestion
interface BundleCreationInput {
```

Following our code conventions.

src/memberBundle.ts
```diff
@@ -0,0 +1,175 @@
...
+  let dnssecChain;
+  try {
+    dnssecChain = await retrieveVeraDnssecChain(orgName);
+  } catch (err) {
+    logger.warn(
+      {
+        memberPublicKeyId,
+        err,
+      },
+      'Failed to retrieve dnssec chain',
```
gnarea, line 70:
```suggestion
      'Failed to retrieve DNSSEC chain',
```

src/memberBundle.ts
```diff
@@ -0,0 +1,175 @@
...
+    memberName,
+    memberCryptoPublicKey,
+    orgCertificate,
+    orgPrivateKey,
+    expiryDate,
+  );
+
+  let dnssecChain;
+  try {
+    dnssecChain = await retrieveVeraDnssecChain(orgName);
```
gnarea, line 63: Sorry, I missed this in the review I did last night:

We should be retrieving the DNSSEC chain before we try to do anything with the KMS: The DNSSEC chain retrieval is expensive CPU- and bandwidth-wise, as well as likely to fail due to the **many** HTTP requests it makes behind the scenes.

So we shouldn't try to use the KMS until we actually have a DNSSEC chain: KMS operations are expensive -- financially and computationally.

Please also write a test for this so we won't introduce a regression in the future.

TL;DR: Move `retrieveVeraDnssecChain()` to the top of this function.

gnarea (changes requested):
Nearly there

src/memberBundle.ts
levansuper: This rework will take place in a dedicated PR. #128

src/backgroundQueue/server.spec.ts
```diff
@@ -51,5 +56,47 @@ describe('makeQueueServer', () => {
...
     });
+
+    describe('makeQueueServerPlugin', () => {
+      const setEnvVars = configureMockEnvVars(REQUIRED_ENV_VARS);
+      const mockFastify: FastifyInstance = {
+        addContentTypeParser: jest.fn(),
+        getDefaultJsonParser: jest.fn(),
+        get: jest.fn(),
+        post: jest.fn(),
+      } as any;
```
gnarea, line 67: Why don't we borrow the approach used elsewhere in this repo? We used all of those methods in other places and never had to mock Fastify to do so.

src/backgroundQueue/server.spec.ts
```diff
@@ -51,5 +56,47 @@ describe('makeQueueServer', () => {
       expect(response.statusCode).toBe(HTTP_STATUS_CODES.BAD_REQUEST);
       expect(response.json()).toHaveProperty('type', QueueProblemType.UNSUPPORTED_EVENT);
     });
+
+    describe('makeQueueServerPlugin', () => {
+      const setEnvVars = configureMockEnvVars(REQUIRED_ENV_VARS);
```
gnarea, line 61: Note that this will create a second mock of the env vars -- each with its own set of `beforeAll()`s and `beforeEach()`s. That may lead to race conditions that may be hard to debug.

We have `getTestServerFixture`. Let's get the mocker from there.

src/backgroundQueue/sinks/memberBundleRequest.sink.spec.ts
```diff
@@ -144,7 +146,7 @@ describe('memberBundleIssuance', () => {
         expect(mockPostToAwala).toHaveBeenCalledOnceWith(
           expect.anything(),
           expect.anything(),
-          new URL(envVars.AWALA_MIDDLEWARE_ENDPOINT!),
+          new URL(REQUIRED_ENV_VARS.AWALA_MIDDLEWARE_ENDPOINT),
```
gnarea, line 149: We don't need the whole `REQUIRED_ENV_VARS` and for consistency with other tests we should simply do:

```suggestion
          new URL(AWALA_MIDDLEWARE_ENDPOINT),
```

gnarea (changes requested):

src/awala.ts
levansuper: This was moved to memberBundle.ts - https://github.com/relaycorp/veraid-authority/pull/119/files#diff-5006ad038ccf0df13d5fba8db8001b0fad11d2deb7c5c40fb830dc68e101ab0bR78

src/backgroundQueue/sinks/memberBundleRequest.sink.ts
```diff
@@ -0,0 +1,62 @@
...
+  event: CloudEvent<unknown>,
+  options: SinkOptions,
+): Promise<void> {
+  options.logger.debug({ eventId: event.id }, 'Starting member bundle request trigger');
+
+  const validatedData = validateMessage(event.data, MEMBER_BUNDLE_REQUEST_PAYLOAD);
+  if (typeof validatedData === 'string') {
+    options.logger.info(
+      { eventId: event.id, validationError: validatedData },
+      'Malformed event data',
```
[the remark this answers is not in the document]
levansuper: missed this

src/backgroundQueue/server.spec.ts
levansuper: Took an example from here.
https://github.com/relaycorp/veraid-authority/blob/main/src/utilities/fastify/plugins/jwksAuthentication.spec.ts

gnarea: But we're testing routes here.

gnarea: Mocking Fastify should only be done in exceptional circumstances (e.g., plugins), not when testing routes.

gnarea (approved):
Nice one, thanks!

PR #129, opened 2023-04-26
test: Implement functional test suite

### LINKED ISSUE #91: Write functional test suite
We have a fairly reliable unit test suite, so this is just to verify that everything works end-to-end, especially our use of backing services (e.g., auth server, Knative eventing). So we should only test the following scenarios:

- Org member creation.
- Org member key import.
- Member bundle request via Awala endpoint (optional: we may want to defer this until the Awala endpoint middleware is implemented).
- Org auth:
  - Anonymous request.
  - Authenticated user with authentication denied.

gnarea:
# Overview

- Implement smoke tests for API server.
- Implement smoke tests for background queue server.
- Implement E2E test where a member claims their bundle via Awala.
- Unfortunately, these tests don't seem to work on macOS: #140

# High-level changes

- 90% of this is dealing with Kubernetes and Knative.
- Had to make some trivial changes to the units under test to facilitate testing (e.g., extract content types).

Fixes #91.


src/testUtils/db.ts
```diff
@@ -36,13 +36,13 @@ export function setUpTestDbConnection(): () => Connection {
...
   beforeEach(async () => {
-    if (connection.readyState === 0) {
+    if (connection.readyState === STATES.disconnected) {
       connection = await connect();
     }
   });
 
   afterEach(async () => {
-    if (connection.readyState === 0) {
+    if (connection.readyState === STATES.disconnected) {
```
gnarea, line 45: Irrelevant to this PR, but I need the latest ESLint config here and it requires these changes.

Issue #131, opened 2023-04-27
Deleting member public key leaves orphan records


gnarea:
I.e., member bundle request.

PR #133, opened 2023-04-27
fix: Delete member bundle request on key deletion

levansuper:
fixes: #131

### DIFF
--- a/src/memberPublicKey.ts
+++ b/src/memberPublicKey.ts
@@ -66,8 +66,8 @@ export async function deleteMemberPublicKey(
 
   // Member bundle request should be deleted before the member public key
   await memberBundleRequestModel.deleteOne({
-    publicKeyId: publicKeyId
-  })
+    publicKeyId,
+  });
   await memberPublicKey.findByIdAndDelete(publicKeyId);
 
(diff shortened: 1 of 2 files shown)


src/memberPublicKey.ts
```diff
@@ -56,10 +57,17 @@ export async function deleteMemberPublicKey(
...
   options: ServiceOptions,
 ): Promise<Result<undefined, MemberPublicKeyProblemType>> {
+  const memberBundleRequestModel = getModelForClass(MemberBundleRequestModelSchema, {
+    existingConnection: options.dbConnection,
+  });
   const memberPublicKey = getModelForClass(MemberPublicKeyModelSchema, {
     existingConnection: options.dbConnection,
   });
 
+  // Member bundle request should be deleted before the member public key
```
gnarea, line 67: This is just describing the code in its current form. It's not explaining why, which is the important part.

```suggestion
  // Defer the key deletion until the end to make retries possible, in case we fail to delete dependant records.
```

gnarea (changes requested):
Just one minor thing found

gnarea (approved):

PR #136, opened 2023-04-27
perf: Add member id to bundle request model

levansuper:


src/awala.spec.ts
```diff
@@ -34,206 +44,243 @@ describe('awala', () => {
...
 
-      await createMemberBundleRequest(methodInput, serviceOptions);
+    const countResult = await memberBundleRequestModel.count({
+      publicKeyId: memberPublicKey._id.toString(),
+    });
+    expect(countResult).toBe(1);
+  });
 
-      const countResult = await memberBundleRequestModel.count({
+  test('Invalid public key id should be refused', async () => {
```
gnarea, line 125: "invalid" is too ambiguous.

```suggestion
  test('Unknown public key id should be refused', async () => {
```

src/awala.spec.ts
```diff
@@ -34,206 +44,243 @@ describe('awala', () => {
...
-      });
-      expect(countResult).toBe(1);
+      },
+      serviceOptions,
+    );
+
+    expect(result.didSucceed).not.toBeTrue();
+  });
+
+  test('Invalid public key Id should not create member bundle request', async () => {
```
gnarea, line 137: Looking at the other tests in this file, wouldn't it be more consistent to bundle this together with the test above? See:

https://github.com/relaycorp/veraid-authority/blob/291f81f7bbf8a4265341290a8e800642aba5a38c/src/awala.spec.ts#L52

src/awala.ts
```diff
@@ -1,20 +1,33 @@
...
   options: ServiceOptions,
-): Promise<SuccessfulResult<undefined>> {
+): Promise<Result<undefined, undefined>> {
   const memberBundleRequestModel = getModelForClass(MemberBundleRequestModelSchema, {
     existingConnection: options.dbConnection,
   });
+  const memberPublicKeyModel = getModelForClass(MemberPublicKeyModelSchema, {
+    existingConnection: options.dbConnection,
+  });
+  const publicKey = await memberPublicKeyModel.findById(requestData.publicKeyId);
```
gnarea, line 22: We don't need the record, right? Let's do an _exists_ query

gnarea (changes requested):

src/awala.ts
levansuper: actually, we need this to get the member Id
https://github.com/relaycorp/veraid-authority/pull/136/files#diff-725e18d8b08af438b9930029c9d9f1d74127beb38f17a39bc726315266bd435cR40

src/awala.spec.ts
gnarea: That works too

gnarea (approved):
Thanks!

PR #138, opened 2023-04-28
feat: Restrict org deletion with existing members

### LINKED ISSUE #123: Prevent org deletion when it has one or more members
We're currently leaving orphan records behind, but instead of manually cascading the delete operation (which in itself is a major undertaking with MongoDB due to the lack of foreign key constraints), we should just prevent its deletion.

So org admins should first manually delete each member, which won't be easy until #122 is implemented.

levansuper:
closes #123


src/org.ts
```diff
@@ -27,6 +29,40 @@ function validateOrgData(
...
+  orgName: string,
+  options: ServiceOptions,
+): Promise<Result<undefined, OrgProblemType>> {
+  const memberModel = getModelForClass(MemberModelSchema, {
+    existingConnection: options.dbConnection,
+  });
+
+  const memberCount = await memberModel.count({ orgName });
+  if (memberCount > 1) {
+    options.logger.info({ orgName }, 'Refused deletion - existing org members');
```
gnarea, line 42: This style isn't consistent with our logs. I'd suggest something like:

```suggestion
    options.logger.info({ orgName }, 'Refused org deletion because it contains multiple members');
```

src/org.ts
```diff
@@ -27,6 +29,40 @@ function validateOrgData(
...
+    return {
+      didSucceed: false,
+      reason: OrgProblemType.EXISTING_MEMBERS,
+    };
+  }
+
+  if (memberCount === 1) {
+    const lastAdmin = await memberModel.findOne({ orgName, role: Role.ORG_ADMIN });
+    if (lastAdmin === null) {
+      options.logger.info({ orgName }, 'Refused deletion - last member not org admin');
```
gnarea, line 52: Similar to the above:

```suggestion
      options.logger.info({ orgName }, 'Refused org deletion because last member is not admin');
```

gnarea (changes requested):
Just a couple of nitpicks

PR #139, opened 2023-04-28
feat: Disable Awala routes if env variable is not set

### LINKED ISSUE #135: Make Awala Middleware optional
Make the respective env var optional and skip the registration of `/awala` routes in that case.

levansuper:
closes #135

### DIFF
--- a/src/api/server.spec.ts
+++ b/src/api/server.spec.ts
@@ -53,7 +53,7 @@ describe('makeApiServer', () => {
       expect(mockFastify.register).toHaveBeenCalledWith(mockRegisterAwalaRoute);
     });
 
-    test('Missing env variable should not register awala routes', async () => {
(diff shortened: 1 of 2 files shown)


src/api/server.spec.ts
```diff
@@ -29,4 +41,24 @@ describe('makeApiServer', () => {
 
     expect(serverInstance).toBe(mockFastify);
   });
+
+  describe('makeApiServerPlugin', () => {
+    const mockEnvVariables = configureMockEnvVars();
+
+    test('Existing env variable should register awala routes', async () => {
```
gnarea, line 48: I think this title is too vague. Suggestion:

```suggestion
    test('Awala routes should be registered if middleware is set', async () => {
```

gnarea (changes requested):
Thanks! All good, just two nitpicks.

gnarea (approved):

PR #152, opened 2023-05-01
fix: Fix namings in logs

### LINKED ISSUE #120: Common logging fields use inconsistent names
- Org name: `orgName`
- Member id: `memberId`
- Public key id: `publicKeyId`

... instead of just `name` or `id`.

levansuper:
fixes #120 

Also changing:
publicKeyId => memberPublicKeyId
token => memberKeyImportToken


gnarea (changes requested):
LGTM, only thing is that the "id" in `memberKeyImportTokenId` is redundant because the token is the id.

gnarea (approved):
Thanks!

PR #182, opened 2023-06-09
feat: Replace structured events with binary ones

### LINKED ISSUE #154: Adapt to final version of the Awala Endpoint
As of this writing:

- [x] Move `POST /awala` to its own CloudEvents server at `POST /`.
- [x] Use **binary** mode instead of structured.
- [x] Replace processing of Awala PDA with use of sender's address (available in the `source` of the CloudEvent received from the member via Awala).
- [x] Remove `AWALA_MIDDLEWARE_ENDPOINT` env var.

levansuper:
closes #154


src/awala/routes/awala.routes.spec.ts
```diff
@@ -71,49 +91,81 @@ describe('awala routes', () => {
...
+
       expect(response).toHaveProperty('statusCode', HTTP_STATUS_CODES.ACCEPTED);
-      expect(mockCreateMemberBundleRequest).toHaveBeenCalledOnceWith(validPayload, {
-        logger: server.log,
-        dbConnection: server.mongoose,
-      });
+      expect(mockCreateMemberBundleRequest).toHaveBeenCalledOnceWith(
+        { ...validMessageContent, peerId: AWALA_PEER_ID },
+        {
+          logger: expect.anything(),
```
levansuper, line 151: The problem with this is that we are passing a child instance of the logger to the method. Not sure how to validate that

src/utilities/buffer.spec.ts
```diff
@@ -11,3 +11,25 @@ describe('bufferToArrayBuffer', () => {
...
+    };
+    const jsonString = JSON.stringify(json);
+    const buffer = Buffer.from(jsonString);
+
+    const result = bufferToJson(buffer);
+
+    expect(result).toStrictEqual(json);
+  });
+
+  test('Invalid json should return null', () => {
```
gnarea, line 28:
```suggestion
  test('Invalid JSON should return null', () => {
```

src/utilities/eventing/Emitter.spec.ts
```diff
@@ -6,11 +6,18 @@ import { configureMockEnvVars } from '../../testUtils/envVars.js';
 import { mockSpy } from '../../testUtils/jest.js';
 import { CE_ID, CE_SOURCE, K_SINK } from '../../testUtils/eventing/stubs.js';
 
+enum CloudEventMode {
+  BINARY = 'binary',
+  STRUCTURED = 'structured',
+  BATCH = 'batch',
```
gnarea, line 12: We could skip those two for maintainability reasons. They could be dropped or more could be added in the future, and we don't want to maintain that.

src/backgroundQueue/sinks/memberBundleRequest.sink.spec.ts
```diff
@@ -60,15 +60,17 @@ describe('memberBundleIssuance', () => {
...
     id: MEMBER_PUBLIC_KEY_MONGO_ID,
     source: CE_SOURCE,
     type: BUNDLE_REQUEST_TYPE,
-
-    data: {
-      peerId: AWALA_PEER_ID,
-      publicKeyId: MEMBER_PUBLIC_KEY_MONGO_ID,
-    },
+    subject: 'https://relaycorp.tech/awala-endpoint-internet',
+    data: JSON.stringify(validMessageContent),
```
gnarea, line 73:
```suggestion
    data_base64: JSON.stringify(validMessageContent),
```

src/awala/routes/awala.routes.spec.ts
gnarea: Indeed. It's not impossible... but the solution is really cumbersome. So let's leave this as it is.

src/awala/routes/awala.routes.ts
```diff
@@ -82,39 +113,52 @@ export default function registerRoutes(
...
-
       const processor = awalaEventToProcessor[contentType!];
 
-      const didSucceed = await processor(request.body, serviceOptions);
+      const message: Message = { headers: request.headers, body: request.body };
+      let event;
+      try {
+        event = HTTP.toEvent(message) as CloudEventV1<unknown>;
+      } catch {
+        return reply.status(HTTP_STATUS_CODES.BAD_REQUEST).send();
```
gnarea, line 140: Let's log what happened.

gnarea (changes requested):
Thanks! Just a few suggestions

src/backgroundQueue/sinks/memberBundleRequest.sink.spec.ts
levansuper: We are testing that the same way in awala endpoint project. "data" generates data_base64

src/awala/routes/awala.routes.ts
levansuper: We are not logging that in awala endpoint project. Should we here?

gnarea: I missed that there. I'll create an issue.

Yes, let's please log it here.

gnarea (approved):

